PR #5, opened 2015-08-11
Added .gitignore files for modified files during linux build

stuckless:
UPDATED README to include instructions for Ubuntu.

Also added .gitignore files for modified files during linux build.  Ideally after a build, there should be no modified files that might accidentally get checked into git.  Hopefully the .gitignores are not TOO restrictive.  I tried my best carefully only select newly added files/directories.


Narflex:
Can you be sure you've signed the Google CLA as outlined in the CONTRIBUTING file? I just checked with your GitHub username and it didn't indicate you've signed it already.

stuckless:
I have signed the CLA.

After doing a rebuild, I had over 500 new files :(

I did start with a bunch of basic common templates for c/c++/java/eclipse and then I had about 250 files of which I added to gitignore anything that was new.  I didn't see any point in storing a binary by-product into git, if that binary is built from sources.

My main goal to clean up my repo so that I can easily manage changed/new files without having to wade through the 500+ other files.

How do you deal with these files?  I was actually surprised to see so many files showing up in the git changes after doing a build?

Narflex:
There's a 'clean.sh' script in there which should cleanup the majority of changes. Although I just checked that and there are still things left around...that script could just get fixed as well to make things OK. :)

PR #7, opened 2015-08-13
Update Singleton creation pattern to be thread safe

TuscanRaider:
I believe I found all of the Singletons that needed tweaking.  Let me know if I missed any.

### DIFF
--- a/java/sage/Carny.java
+++ b/java/sage/Carny.java
@@ -70,18 +70,20 @@ public final class Carny implements Runnable
     else return "Unknown";
   }
 
-    public static Carny prime() {
-        return getInstance();
-    }
+  public static Carny prime()
+  {
+    return getInstance();
+  }
 
-    private static class CarnyHelper {
+  private static class CarnyHelper
+  {
+    public static final Carny instance = new Carny();
+  }
 
-        public static final Carny instance = new Carny();
-    }
-
-    public static Carny getInstance() {
-        return CarnyHelper.instance;
-    }
+  public static Carny getInstance()
+  {
+    return CarnyHelper.instance;
+  }
 
   private Carny()
   {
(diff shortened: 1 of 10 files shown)


ai7:
Nice! You may want to change the indent size (4) to match the current one used in the file (2).

Narflex:
+1 to fixing indent sizes :)

I do have an issue with how this was done. It's changing the timing of when all of these singletons are getting initialized. Now they are will be initialized when the class is initialized...not when the first call to getInstance() is made. This will likely have unintended consequences and will probably even crash when run. For example, if the class loader decides to initialize MMCHolder before the Sage class loads the property system..it will fail to be able to read the properties.

Can you just do it in a way where the logic is the same but fixes that one highly unlikely race condition mentioned in the article? (about the var getting assigned before the object is fully constructed)

TuscanRaider:
I have addressed the indenting.  I attempted to match the existing style with the braces too, though that's not consistent :)  I suppose you can reject the pull request and I will resubmit. 

With regard to the initialization timing, the objects are not created until referenced through the getInstance() method (internal classes are not loaded until referenced).  The object timing should be consistent with the previous implementation.

Ref: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom

Narflex:
OK, cool...that satisfies me and is a pretty nifty technique. :) One other change then...can you modify CarnyHelper to instead by CarnyHolder so we have consistent naming with this technique?

TuscanRaider:
Cool. Ooops, yes I meant to fix that.  I am ready to recreate the pull request.

Narflex:
Don't forget to add yourself to the CONTRIBUTORS and AUTHORS files. :)

PR #17, opened 2015-08-17
Clean up build files

jreicheneker:
Removes all files that show up in 'git status' after a build.
- Changed several clean targets to distclean
- It has to manually clean up some files in faac and faad2
- Fixed a couple wrong filenames in Makefile clean targets
- Clean up files in native/lib/TSNative and native/lib/NativeCore
- Change a few Makefiles in native/ax/


Narflex:
Thanks Jason. :)

PR #19, opened 2015-08-18
Fix for 64-bit scan crash

jason-bean:
Changed crc from unsigned long to uint32_t to fix crash on 64-bit systems.


TuscanRaider:
It woukd be good to update the return type and the header to match the uint32_t type

Issue #2, opened 2015-08-10
No specific User Agent for SageTV


CraziFuzzy:
When downloading content (or anything) using sage.FileDownloader, no user_agent is specifically set, so it uses the default "Java/x.x.x_xx" user_agent.  It seems to be more 'correct', it should be setting a user_agent that would be more like "SageTV/x.x.x".  This should be able to be done by either:
1.  Adding it directly into FileDownloader, which would require some logic to build the string from the current build version.  This is the quickest way to implement this.  The downside is it would not allow a way to 'force' the user agent to something else, which might be required when trying to access some sources.
2.  Using the already implemented requestProps argument in FileDownloader to pass the user agent from the calling library (usually Global.StartFileDownload).  This would then likely require adding an overloaded argument to the StartFileDownload function to allow the API call to set the User Agent, which if NOT set would default to "SageTV/x.x.x"

The immediate reason I am wanting this implemented is so that PlayOn can recognize SageTV, so SageTV can be white-listed as an HD capable player, which is the only way PlayOn will stream content in HD.  Option 2 would allow sage to emulate an already white-listed player to get PlayOn in HD immediately.  Option 1 would require PlayOn to have SageTV added to it's white-list before it would work in HD.

Narflex:
One thing I'm not following....is can't this already be done using the StartFileDownload API call?  It sounds like #2 is already implemented. Take a look at the overloaded StartFileDownload API call in sage/api/Global.java. IIRC, that's why I added that argument in the first place. :)

And the main reason this was never set as the default was I didn't want SageTV getting blacklisted from specific sites.

CraziFuzzy:
At a glance, it didn't look like StartFileDownload actually passed through anything but ServerAddress, SourceFile, and DestFile when it calls downloadFile.

CraziFuzzy:
On another look, it looks like StartBackgroundFileDownload handles the props...

CraziFuzzy:
but StartFileDownload and StartCircularFileDownload do not.

Narflex:
Ahh yes, you are correct. :)

OK, I'd propose making 2 changes then.

1. Have a property such as "use_sagetv_user_agent" which then uses "SageTVa.b.c" as the user agent (and it would default to not being on for backwards compatibility)

2. Add an overloaded StartFileDownload API call which takes a RequestProperties parameter so this can force the user agent to whatever it wants.

That all sound good?

CraziFuzzy:
[...]

PR #20, opened 2015-08-18
Download request properties

CraziFuzzy:
Added ability to specify request header properties specific to an individual file download request.  Primarily done to set a specific user-agent that might be required for a given file.

This should close out issue #2.

### DIFF
--- a/java/sage/api/Global.java
+++ b/java/sage/api/Global.java
@@ -2106,7 +2106,7 @@ public class Global {
         else
           return FileTransfer.getFileTransfer(stack.getUIMgrSafe()).getStatusMessage();
       }});
-    rft.put(new PredefinedJEPFunction("Global", "StartFileDownload", new String[] { "ServerAddress", "SourceFile", "DestFile" })
+    rft.put(new PredefinedJEPFunction("Global", "StartFileDownload", -1, new String[] { "ServerAddress", "SourceFile", "DestFile" })
     {
       /**
        * Instructs the file transfer engine to download the specified file from the server to the local destination file. You may also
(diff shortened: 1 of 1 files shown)


java/sage/api/Global.java
```diff
@@ -2171,11 +2190,32 @@ public Object runSafely(Catbert.FastStack stack) throws Exception{
...
+       */
       public Object runSafely(Catbert.FastStack stack) throws Exception{
+        java.util.Properties props = null;
+        if (curNumberOfParameters == 4)
+          props = (java.util.Properties) stack.pop();
         java.io.File dest = getFile(stack);
         String src = getString(stack);
         String serverName = getString(stack);
-        return FileDownloader.getFileDownloader(stack.getUIMgrSafe()).downloadCircularFile(serverName, src, dest);
+        return new FileDownloader(null).downloadFile(serverName, src, dest, true, props);
```
Narflex: Why did you change this to construct a new object rather than using the one associated with the UIManager? This method is for foreground downloading which is used during playback, so it needs to use the one linked with that UIManager or it'll break other thing.

CraziFuzzy: Honestly, I don't have an answer.  If anything, It looks like a copy/paste error, pulling from the background downloader on accident.  I likely didn't notice the problem because I don't have anything currently that uses the circular download.

Oh, just remembered why I was copying and pasting in the first place.  downloadCircularFile doesn't have an overload that brings in the props, so I went with downloadFile and set the circular boolean to true - but no, it shouldn't be creating a new object, so I'll get that fixed up and make a new pull request.

PR #22, opened 2015-08-19
Trim string created from reading activkey file

jason-bean:
Whitespace characters were not being stripped when reading activkey file resulting in an inability to connect to EPG server.


Narflex:
Thanks for fixing my bug. :)

PR #27, opened 2015-08-25
update libhdhomerun to 20150615

webernissle:
I updated the libhdhomerun to the latest version. The only change for HDHomeRun.cp was the version check. I see this as a prerequisite in order to further improve some features with HDHomeRun on linux. I discovered issue with an older EU DUAL model using DVB-C (sage assumes it is only DVB-T), with my new hdhomerun3_dvbc I have 4 tuners but sage sees only 2 of them, I think there is a hard coded limit of tuner set to 2 in HDHomeRun.cp. Will do further work on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ config*.log
 /bin/
 /.classpath
 /.project
+/.cproject
 /.gradle/
 /.settings/
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -21,3 +21,4 @@ Christopher Piper <fuzzy@weirdness.com>
 Jason Reicheneker <jason.reicheneker@gmail.com>
 Raymond Chi <raychi@gmail.com>
 Joshua Lewis <josh@joshandmonique.com>
+Martin Weber Nissle <webernissle@gmail.com>
--- a/java/sage/WarlockRipper.java
+++ b/java/sage/WarlockRipper.java
@@ -245,7 +245,7 @@ public class WarlockRipper extends EPGDataSource
   }
 
   // This may take some time due to the diskspace calculation so do it before we connect to the server
-  public static String getSubmitInfo(Wizard wiz, String providerID)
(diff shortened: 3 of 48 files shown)


Narflex:
Before I accept this change...I'd want to get confirmation from someone using the HDHomeRun in the US under both ATSC and QAM that they both still work.  Can you ask on the forums for some people to try it out?

webernissle:
Created the forum entry asking for further tests
http://forums.sagetv.com/forums/showthread.php?t=62380

Narflex:
If you don't get feedback after a few days I'll just go ahead and merge it...then if others have problems later, they can try with/without this change. Although I'd probably wait until after the 64 bit issue is fixed (if you don't get responses to it being tested) since I know that's currently being worked on and I don't want to further complicate that issue. :)

Narflex:
If you don't get feedback after a few days I'll just go ahead and merge it...then if others have problems later, they can try with/without this change. Although I'd probably wait until after the 64 bit issue is fixed (if you don't get responses to it being tested) since I know that's currently being worked on and I don't want to further complicate that issue. :)

Narflex:
If you don't get feedback after a few days I'll just go ahead and merge it...then if others have problems later, they can try with/without this change. Although I'd probably wait until after the 64 bit issue is fixed (if you don't get responses to it being tested) since I know that's currently being worked on and I don't want to further complicate that issue. :)

Narflex:
We good to merge this one now? (just rebase and resolve the conflicts first, then I can one-click merge it)

webernissle:
Only the Contributors file showed a conflict, I resolved it with 94b1c7d.  "Taddeusz" tested it on ATSC & QAM in US.

webernissle:
I just saw that my other commit is now part of this request (4a3d578) which is a first success to get the lib HDHomeRun build on MAC OS/X. Everything without the X11 part should work or possible to replace by an equivalent OS/X library. But you can just merge in the libhdhomerun upgrade.

PR #31, opened 2015-09-03
Log full stacktrace on EPG plugin errors

Slugger:
EPG thread only logs the Throwable.toString() when it catches an
exception from an EPG plugin, which isn't very helpful when tracking
down bugs.  A full stacktrace makes solving problems rather easy.

### DIFF
--- a/java/sage/EPG.java
+++ b/java/sage/EPG.java
@@ -15,7 +15,10 @@
  */
 package sage;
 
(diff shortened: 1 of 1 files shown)


Narflex:
Can you change it to use Sage.printStackTrace(Throwable) instead? :)

Slugger:
Branch now using Sage.printStackTrace() instead.

PR #32, opened 2015-09-04
Got tired of seeing the generated files in my git status

stuckless:
the files are explicitly ignored, as to not accidentally include other files... and all files listed are there after a build, but are gone during a clean, so they are all build "by product" files.  I really don't like to see build by product files showing up in my git status :)


stuckless:
The JTUX stuff got tacked on on here... it thought github would create 2 separate pull requests, but the JTUX part allows for 64bit testing... You can see the comments in the JTUX code.

CraziFuzzy:
Yeah, I've fought github on trying to narrow down a pull request to just select commits.  As far as I can tell, the only way to do that is to have just the commits you want in a separate branch, and then pull from there.

stuckless:
Closed it, since these changes are in the mplayer pull request as well.

PR #33, opened 2015-09-04
mplayer upgrade

stuckless:
This appears to being in my previous pull request as well.

There are lots build changes for this... you'll notice most of the build scripts are updated.  In some cases the linux scripts simply call the gradle task.  The reason is that jogl is a dependency and gradle will download it on the fly, as needed.  (Also makes it easy to switch jogl versions)

Mplayer will download the opensagetv site that hosts the mplayer fork with changes.  It will be configured and built using the remote sources.

There are losts of little gradle tweaks as well, including a task that will publish to bintray if the API key is provided.


jreicheneker:
The overall build doesn't fail if the Sage.jar or MiniClient.jar builds fail. In my case, IntelliJ had a lock on build.gradle and caused the calls to gradlew to fail.

It can be reproduced by changing a Java source file and causing it to have a build failure.

stuckless:
This is when you use buildall.sh?  I don't know why having a lock on the build.gradle would cause gradle to fail... this must be a windows-ism.  Any ideas on how we can fix it?

jreicheneker:
I'm on Linux :) I don't know about fixing the lock, but the build should fail if the gradlew calls fail.

stuckless:
That's good to know... I should be able to reproduce the lock then :)  I'm
more interested as to why the build.gradle is locked and why it would
prevent gradle from doing a build.

But, I'll add a check to fail the build if the jar fails.

jreicheneker:
I think it's happening if I launch SageTV server with the linuxRunServer task in the IntelliJ Gradle tool window and leave it running when starting buildall.sh.

stuckless:
I've updated to build script to return the gradle return code, so if gradle fails, then the build fails (as it should).   The specific locking issue is a gradle bug (or feature) that they have recognized, and they might fix it.  ie, you can't have 2 concurrent "builds" running at the same time for the same codebase.  This is a feature, in that, doing so could corrupt the build, but in these cases, the tasks are not build tasks, so there'd be no reason to worry.   

We could change the launch task to use ant (as per this thread https://discuss.gradle.org/t/does-javaexec-task-support-spawning-the-jvm-in-a-new-process-which-will-outlive-gradle/511/5 ) which does support forking, so it would not hold the process when launching.

jreicheneker:
That issue is about 5 years old, I'm not holding my breath for them to fix it :)

It's not a big issue to me here because I usually need to stop the processes anyway if I'm rebuilding. Long term I'd like to enhance the build.gradle to integrate with IntelliJ compilation and run configurations better. I haven't used Gradle much and there are a lot of other things higher on my list so I don't see myself working on it anytime soon.

build/buildmplayer.sh
```diff
@@ -0,0 +1,37 @@
...
+if [ "$MPLAYER_NEW" = "1" ] ; then
+	mkdir tmp/
+	cd tmp/
+	rm -rf mplayer
+	git clone -b sagetv https://github.com/OpenSageTV/mplayer.git
+	cd mplayer
+	echo | ./configure-sagetv
+	make -j32
+	echo "Built NEW mplayer"
+	cp -v mplayer ../../elf/mplayer
```
Narflex: The elf directory is created after this script is run...so this copy command will end up failing.

build/buildmplayer.sh
```diff
@@ -0,0 +1,37 @@
...
+	cd ../../
+else
+	# use legacy mplayer build
+	cd ../third_party/mplayer/
+	EXTRA_CFLAGS="-fno-common -DMINGW_MEMALIGN=1" ./configure --host-cc=gcc --enable-runtime-cpudetection --disable-mencoder --disable-gl --enable-directx --enable-largefiles --disable-langinfo --disable-tv --disable-dvdread --disable-dvdread-internal --disable-menu --disable-libdvdcss-internal --enable-pthreads --disable-debug --disable-liba52 --disable-freetype --disable-fontconfig --enable-stv --enable-stream-sagetv --disable-ivtv --disable-x264 --extra-libs=-lpthread || { echo "Build failed, exiting."; exit 1; }
+	make -j32 || { echo "Build failed, exiting."; exit 1; }
+	cp -v mplayer ../../build/elf
```
Narflex: The elf directory is created after this script is run...so this copy command will end up failing.

build/buildsage.sh
```diff
@@ -15,40 +15,10 @@
...
-rm -rf images
-mkdir images
-cd images
-cp ../../../images/SageTV/images/*.gif .
-cp ../../../images/SageTV/images/*.png .
-mkdir studio
-cd studio
-cp ../../../../images/SageTV/images/studio/*.gif .
+
+# now uses gradle because there a dynamic dependencies for JOGL
```
Narflex: That shouldn't be true...there should be no JOGL dependencies for the Sage.jar file.

java/sage/miniclient/MiniMPlayerPlugin.java
```diff
@@ -544,7 +561,16 @@ else if (hostname != null && hostname.length() > 0)
         else
           cmds.add(new java.io.File(file).getPath());
 
-        if ("TRUE".equals(MiniClient.myProperties.getProperty("player_cmdline_debug", null))) System.out.println("Executing mplayer cmd: " + cmds);
+        String args[] = (String[])cmds.toArray(new String[0]);
```
Narflex: Now that you're creating an args[], you can reuse it when launching mplayer below. :)

build/clientfiles/sageclient.sh
```diff
@@ -29,6 +29,6 @@ echo Changing to SageTV directory $sagepath
 cd $sagepath
 
 export LD_LIBRARY_PATH=.
-java -cp MiniClient.jar:gluegen-rt.jar:jogl.all.jar:nativewindow.all.jar:. sage.miniclient.MiniClient $1 $2 > /dev/null 2>&1
-#java -cp MiniClient.jar:gluegen-rt.jar:jogl.all.jar:nativewindow.all.jar:. sage.miniclient.MiniClient $1 $2 > miniclient.log 2>&1
+#java -cp `ls -1 *.jar | tr '\n' ':'`:. sage.miniclient.MiniClient $1 $2 > /dev/null 2>&1
+java -cp `ls -1 *.jar | tr '\n' ':'`:. sage.miniclient.MiniClient $1 $2 > /tmp/miniclient.log 2>&1
```
Narflex: Many people override the behavior of 'ls' in their linux configs. I'd rather you use the same technique in startsagecore for determining what the JAR files are so we're consistent between the different scripts.

Narflex:
And in the future...please break up stuff like this into multiple check-ins please. :)

build/buildall.sh
```diff
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+# NOTE Passing MPLAYER_NEW=1 when calling this script 
+# will cause the newer mplayer to be built
+export MPLAYER_NEW=1
```
Narflex: Can you change this to use ?= or := (I forget which one sets it only if not set already) so that then it can be overridden without editing the build file directly?

Narflex:
I did a test with this, and running it for the first time does not end up with tarballs/deb packages that include mplayer. I'm fairly sure it's the failure of the cp command at one point that's the cause.

Narflex:
Even on a second run...there's still no mplayer included in the client packages.

stuckless:
Jeff, thanks for feedback... I've fixed most of the things you've addressed.

You said you wanted this in separate checkins... This is spread over 9 different commits.  Not sure how much more broken up you want.  Are you saying you'd want 9 different pull requests??

Was just doing some testing as well, and the build script fails to detect failed mplayer build (so I fixed that )... But it does appear to be an inconsistency between ffmpeg (which mplayer auto-pulls) and mplayer core build.  I'm going to start to maintain a separate ffmpeg repo so that this doesn't happen.

Regarding ffmpeg (but not related to these changes)... do you have the original, unmodified, ffmpeg sources that you started with (much like you had with the mplayer sources).

stuckless:
Jeff... I just updated the mplayer sources to the latest, so, you can run another test build, if you want ;)

Narflex:
Doing it like this with commits is fine. I was just always reviewing the pull requests as a whole before...but I can just do it commit by commit a well. :) I'll try this out now.  And I no longer have those FFMPEG sources...see my forum post...my old workstation took a turn for the worst. (at least it did that now and not before I got SageTV open sourced and could verify everything would build on Windows)

Narflex:
Have you tried this remotely so that it uses the transcoder as well? I'm getting problems with it hitting the EOF immediately when playing back that way.

I still need to look at your mplayer diffs themselves.  Do you also happen to have the current mplayer version + patches from SageTV diffs that I can look at?  Then I can compare that to see what changes you ended up not including.

Narflex:
I can confirm that with the OLD mplayer build, that I can play stuff remotely (i.e. transcoded) properly.

Narflex:
And I'm fine accepting this change now if you modify the default to be using the old mplayer. :) And then once those issues get fixed with the new mplayer, then we can change the default.

stuckless:
Jeff, if you go to 
https://github.com/OpenSageTV/mplayer/compare/sagetv
You'll see the changes that I did to the new mplayer.  (the mplayer "master" is original, and the "sagetv" branch is where I keep adding the changes)

To get your original changes, you can look at the 
https://github.com/OpenSageTV/mplayer-sage9orig
repo

When I was merging changes, it was a tedious process... basically I used MELD to do a 2 way compare on the sage9orig sources using a bunch of filters to hide things that are ONLY in the OLD sources by not in yours, etc.  This gave me a short list of changed files.  I then used MELD to do a 3way compare using the sage9orig into the current mplayer, and I went file by file, merging stuff.  I didn't do anything in the ffmpeg area.

If you want, I can probably do a git diff on the 30 files that I modified... might take a couple days.

So, I'll set the default build to build the OLD mplayer and look into the issue that you mentioned.

Narflex:
I took a look at it, and the problem definitely relates to the 'active file' stuff. Not having that set properly would cause the problem where you hit EOF really shortly after starting playback. You can possibly also reproduce this by just playing live TV (without transcoding)...but the stream_sagetv implementation may abstract the problem away there.

When it does placeshifting, it actually downloads the data into a temp file and then just plays that back as a file source (and if you seek or whatever, then it just seeks you to the end of the file and then starts reading the data again from there). Yes, it was done in a very hacky way for that. :) Ideally there'd just be some kind of 'push' mode where you can feed the data into mplayer and it decodes it from that...and then have commands to flush it whenever you do seeks and then start pushing again (which is analogous to how the hardware decoders work on the extenders).  If they have something like this now; it'd be better to convert it over to using that (but I didn't see anything when looking just now).

The active file changes bleed into ffmpeg as well...yet I don't recall for sure if those changes are required in mplayer for it to work properly...but it's very likely that they are.

stuckless:
Thanks Jeff... I'll do some testing and see what I can find.

PR #34, opened 2015-09-09
More gitignore updates

Slugger:
Just a few more gitignore updates that I noticed in my tree.


Slugger:
Branch now addresses all concerns.

Issue #1, opened 2015-08-10
Fire a System Message when the bitrate for a recording is too low or too many segments


Narflex:
One of the few problems I still have on occasion with SageTV is when one of my STBs goes off due to a device upgrade by Verizon; or one of my HDPVRs just starts having issues.  This manifests itself in 1 of 2 ways. 

1. The file size of the recording is much lower (over 50% smaller) than normal because it was just a black screen the whole time.
2. There's LOTS of segments created for the recording.

I should add code in the Seeker which detects when either of these 2 cases occurs and then generates a unique system message. I already suppress the one in my own system for when recordings have breaks in them (since this happens too frequently with the HDPVR); which is why I want a new one for this.

I'd make 2 separate property settings for this since not everyone would want this type of detection. There'd also be a setting for the # of segments to cause this to be hit. I'd also have a property for either a specified rate to fire this at; or just have it go through the recordings for that tuner and then calculate an average rate correlated with resolution and then have a configurable threshold for when this fires based on that average.

CraziFuzzy:
Does the HD-PVR work in VBR mode?  if not, isn't the filesize always going to be as expected even with just a plain black screen?

Narflex:
Fuzzy,

I don't recall if it's VBR or CBR mode...but I know for sure that the file sizes are about 25% of normal when its all black screen (I measure by looking at about a dozen recordings I had done).

stuuk:
Sometimes there is a delay in tuning DVB-T radio in the UK which typically show some kind of static image that changes infrequently. Different behaviour on different extenders (mvp tunes much more quickly than say an HD300) maybe is related?

Narflex:
I've committed one relating to too many halts during a recording, it's here: https://github.com/google/sagetv/commit/1d5f49c06a92657a3e7a4866c9dbff8ddc3bda44

Narflex:
And the other one for bitrate is done too: https://github.com/google/sagetv/commit/980a772b9df0b401389bb42702f85c67646301e2

For the bitrate one; I just made it based on a property setting for the capture device...was much easier to implement that way. Of course after I did this one...I realized that the bitrate one would likely cover both cases..I'll find out next time my HDVPRs get jacked up. :)

PR #35, opened 2015-09-22
Allow for favorites to be disabled

skiingwiz:
* Add a flag to favorites (Agent) that allows for them to be disabled.  
* Add the logic in Carny to test that flag.  
* Add API methods to query and set that flag.

Note: The Agent flag is whether or not the Agent is DISabled, because that allows the default state of the flag (0) to mean ENabled (i.e. disabled is false).  The API methods test and set whether the favorite is ENabled (which is the opposite of the Agent flag) because I thought that made more sense for a user of the API.

I'll be submitting a separate pull request with an STV update to expose this new functionality once this pull request is accepted..  I've tested with a locally modified v7 STV.  I just need to get those changes made to the latest STV from github.

### DIFF
--- a/java/sage/api/FavoriteAPI.java
+++ b/java/sage/api/FavoriteAPI.java
@@ -980,7 +980,7 @@ public class FavoriteAPI {
           Agent a = (Agent) stack.pop();
           if (Permissions.hasPermission(Permissions.PERMISSION_RECORDINGSCHEDULE, stack.getUIMgr()))
           {
-              Carny.getInstance().setAgentFlags(a, Agent.DISABLED_FLAG, b?0:Agent.DISABLED_FLAG);
+              Carny.getInstance().enableFavorite(a, b);
               Carny.getInstance().kick();
           }
           return null;
--- a/java/sage/api/FavoriteAPI.java
+++ b/java/sage/api/FavoriteAPI.java
@@ -113,6 +113,21 @@ public class FavoriteAPI {
       public Object runSafely(Catbert.FastStack stack) throws Exception{
         return Boolean.valueOf(((Agent) stack.pop()).testAgentFlag(Agent.DELETE_AFTER_CONVERT_FLAG));
       }});
+    rft.put(new PredefinedJEPFunction("Favorite", "IsEnabled", 1, new String[] { "Favorite" })
+    {
+      /**
+       * Returns true if SageTV considers this favorite when performing scheduling.
+       * @param Favorite the Favorite object
+       * @return true if this Favorite is enabled, false otherwise
+       * @since 9.0
+       *
+       * @declaration   public boolean IsEnabled(Favorite Favorite);
+       */
+        @Override
+        public Object runSafely(Catbert.FastStack stack) throws Exception{
+            //Note here that the value of the test is negated
+            return Boolean.valueOf(!((Agent) stack.pop()).testAgentFlag(Agent.DISABLED_FLAG));
+        }});
     rft.put(new PredefinedJEPFunction("Favorite", "GetKeepAtMost", 1, new String[] { "Favorite" })
     {
       /**
@@ -948,6 +963,27 @@ public class FavoriteAPI {
         }
         else
           return Boolean.FALSE;
+      }});
+    rft.put(new PredefinedJEPFunction("Favorite", "SetEnabled", 2, new String[] { "Favorite", "Enabled" }, true)
+    {
+      /**
+       * Sets whether or not SageTV will use this favorite when scheduling recordings
+       * @param Favorite the Favorite object
+       * @param Enabled true if this Favorite is to be used for scheduling, false otherwise
+       * @since 9.0
+       *
+       * @declaration   public void SetEnabled(Favorite Favorite, boolean Enabled);
+       */
+      @Override
+      public Object runSafely(Catbert.FastStack stack) throws Exception{
+          boolean b = evalBool(stack.pop());
+          Agent a = (Agent) stack.pop();
+          if (Permissions.hasPermission(Permissions.PERMISSION_RECORDINGSCHEDULE, stack.getUIMgr()))
+          {
+              Carny.getInstance().setAgentFlags(a, Agent.DISABLED_FLAG, b?0:Agent.DISABLED_FLAG);
+              Carny.getInstance().kick();
+          }
+          return null;
       }});
     rft.put(new PredefinedJEPFunction("Favorite", "GetFavoriteForAiring", 1, new String[] { "Airing" })
     {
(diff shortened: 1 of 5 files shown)


java/sage/api/FavoriteAPI.java
```diff
@@ -113,6 +113,21 @@ public Object runSafely(Catbert.FastStack stack) throws Exception{
       public Object runSafely(Catbert.FastStack stack) throws Exception{
         return Boolean.valueOf(((Agent) stack.pop()).testAgentFlag(Agent.DELETE_AFTER_CONVERT_FLAG));
       }});
+    rft.put(new PredefinedJEPFunction("Favorite", "IsEnabled", 1, new String[] { "Favorite" })
```
Narflex: Please use a more descriptive name than IsEnabled. The API call names are global...and while the other Favorite APIs could be more specific...'Enabled' is just too generic. Can you make it IsFavoriteEnabled?

java/sage/api/FavoriteAPI.java
```diff
@@ -949,6 +964,27 @@ else if (slotType == BigBrother.FULL_ALIGN)
         else
           return Boolean.FALSE;
       }});
+    rft.put(new PredefinedJEPFunction("Favorite", "SetEnabled", 2, new String[] { "Favorite", "Enabled" }, true)
```
Narflex: Same thing here...can you make this SetFavoriteEnabled?

Narflex:
This should work fine (please make the changes to the API names though)...but there's going to be one issue associated with this. When the user makes a change to this state....it won't reflect itself anytime soon in the UI. It will take a full cycle of Carny processing in order for this change to show up. On new systems...that'll be minutes...on old systems...it could be a whole day.

It is possible to do this update right away...which is what happens when Carny.update/add/removeFavorite are called.  You'd have to separate out the logic for updating all the state information in Carny in the add/remove functions and put them into their own functions that can then be called from add/remove and then also the appropriate ones when a Favorite is enabled/disabled.

If you'd prefer I make that change, that's fine...it's touching some tricky parts of the code. :)

skiingwiz:
I've made the API name updates.  I have a few questions about the last comment regarding updating immediately vs. a full Carny cycle.  
* I modeled this after things like SetKeepAtMost, which just updates the Agent flags and then kicks Carney.  I guess that means KeepAtMost changes can also take a while to propagate to the recording schedule?
* Would it work to call updateFavorite after updating the flag and then in updateFavorite, if the disabled flag is set, the set the dontScheduleThisAgent variable to true?

When I tested it was on a clean install, so of course everything happened pretty much immediately for me.

If you want to make this change that's fine, I don't mind.  I am trying to become familiar with this section of the code base, but you obviously have more experience with it than anyone else. :)

Narflex:
The other ones like SetKeepAtMost don't really affect things in a way that'll alter the schedule to a degree people would notice. If autodelete is on; then keep at most is enforced by the Seeker. It's only if keep at most is on and autodelete is off; then it will affect scheduling more of them. (and honestly, I don't know if many people ever use keep at most + autodelete off)

With disabling/enabling...then it'll become more obvious because someone would very likely go check to see if things look right after that to see if they're OK.

Calling updateFavorite won't work because it returns early if the Favorite hasn't changed according to the AGENT_SORTER (and it won't have in this case because that doesn't care about the agentFlags field).

I'm more than happy to help work with you on this yourself so you can learn more (i.e. review your code and give you constructive feedback)...that's the overall goal with the open source project since I already dedicated 10 years of my life to SageTV...it's time for others to have their crack at it. :)

Actually...after looking at this again...it should be pretty easy to do with these couple changes:

1) Add another variant of removeFavorite which is private that has 2 args, and the second one specifies whether the DB should actually be updated. If its true, go on like normal...if its false, then skip the wiz.removeAgent(fav) call (and also fix the debug statement). You could also remove the processing of 'allAgents' in all cases and not grab the DB writeLock if you're not modifying the DB. And then get the 'allFavs' data by calling Wizard.getFavorites() instead (that'll just end up with cleaner code).

2) Add another variant of addFavorite...with a similar goal as #1.

3) Create a method in Carny just for doing enable/disable so it can then call the right thing to do a full update immediately after those changes occur.

I'm OK approving this change the way it is and you doing the rest as another one...or you can roll it all into one. Either way. :)  If you want it approved now, just let me know.

Narflex:
Note to self....check on if Agent.followsTrend should be changed for this
commit too. :-)

Jeff Kardatzke

skiingwiz:
Ok, please go ahead and merge this change in and I'll do the rest as another one.  That will allow me to work on that change as well as the STV update to expose this functionality.

Narflex:
Found another issue based on my 'note to self' from yesterday. Easy fix though. :) You need to update Agent.followsTrend so that it returns false if the Agent is disabled. This call is also used in various places (like add/update/removeFavorite).  The only downside to this is then the Additional Airings for a disabled favorite will return nothing...but I guess that's sort of expected.

skiingwiz:
Ok.  That makes sense.  I updated followTrend to return false for disabled Agents.  I also created a javadoc comment for follow trend.  (I figured I might as well add method comments as long as I'm in there.  Maybe it'll help someone else.)

It would be nice if Additional Airings could still be seen for disabled favorites.  Maybe in the future I can add an overloaded GetFavoriteAirings call that allows the disabled status to be ignored.  Then the STV could call that if the favorite is disabled and still show additional airings. But that is a change for another day.

skiingwiz:
I have the other changes done. I just need to do some testing. So it might be easier for you to just wait a bit and merge it all at once.

Narflex:
OK, I'll wait until all the changes are complete then. But please make the
STV commit a separate commit from the changes to SageTV core.

On Sun, Sep 27, 2015 at 8:41 PM, Chris Danser <notifications@github.com>
wrote:

> I have the other changes done. I just need to do some testing. So it might
> be easier for you to just wait a bit and merge it all at once.
>
> —
> <https://github.com/google/sagetv/pull/35#issuecomment-143632854>.
>



-- 
Jeffrey Kardatzke
jkardatzke@google.com
Google, Inc.

skiingwiz:
Changes are added to this pull request.   They address updating the Carny state immediately, instead of waiting for a reprocess.  Let me know what you think.

Don't worry, STV updates will be in a completely different pull request when I get around to those.

skiingwiz:
Oh, one thing I wanted to note: When a favorite is enabled or disabled, I have Carny fire the FAVORITE_MODIFIED event.  I just wanted to make sure you saw that and don't think it'll cause any issues.  It seemed to me like the right thing to do.

Narflex:
Thanks!

PR #49, opened 2015-10-16
Ubuntu License Key Prompt

jason-bean:
Added debconf features to prompt user for SageTV license key and to create the activkey file when one is provided.

### DIFF
--- a/build/ubuntufiles/server/DEBIAN/config
+++ b/build/ubuntufiles/server/DEBIAN/config
@@ -1,0 +1,11 @@
+#!/bin/sh -e
+. /usr/share/debconf/confmodule
+
+db_input high sagetv-server/haskey || true
+db_go
+
+db_get sagetv-server/haskey
+if [ "$RET" = "true" ]; then
+   db_input high sagetv-server/activkey || true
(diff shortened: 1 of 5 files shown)


stuckless:
Will this always prompt?  ie, what is flow here... I have a couple Vagrant images where I go from source to debs and install the debs... would this stop the flow and ask for key?  Can this be configured to not ask for a key?

jason-bean:
It will only prompt on a clean install or after a purge. Simply uninstalling the sagetv-server package is not enough to trigger the prompts.

stuckless:
So, on a clean install, is there a way to suppress the prompt?  Consider a docker container (or vagrant image) that installs sagetv where there would be no "person" doing the dpkg -i since it is handled as part of a flow.  would I echo "" | dpkg -i or is there some to mechanism so that it doesn't stop and prompt.

jason-bean:
Looks like an option would be to use a preconfiguration file to pre-load "sagetv-server/haskey boolean false" or "sagetv-server/haskey boolean true" and "sagetv-server/activkey string <key>" into the debconf database.

PR #52, opened 2015-10-25
wrapper for comskip

yoavp77:
not sure if you want this sort of thing in github or in a separate plugin repo, but I have some related scripts like this to do a few things (add ac3 tracks to dts-only mkv's so they would work on hd200, etc)

### DIFF
--- a/third_party/comskip/skipcoms.sh
+++ b/third_party/comskip/skipcoms.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+# Script to run comskip against hdpvr recordings. Requires comskip.exe and wine.
+
+# location where recordings are stored
+recordings_location=/recordings/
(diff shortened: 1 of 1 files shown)


Narflex:
This doesn't seem like it would really belong in the SageTV repo unless all the other components it depends on were also in there (like comskip).  It also doesn't seem like it would be a plugin unless it's done in a way that actually gets called by a plugin somehow (since it's a script). Maybe ask on the forums and see what other people say? I'm not really sure what to do about useful things like this. :)

stuckj:
Not sure if you were aware, but there is already a commercial detection plugin (uses comskip pre-packaged) and a corresponding UI mod available in the plugin repo for displaying when comskip is running. I just discovered it myself in getting a Linux sage box up and running. :)

PR #56, opened 2015-11-27
Monitor lost/offline network encoders

CraziFuzzy:
If a network encoder is configured and not found, it was previously ignored, unless the user navigated to the source details page, which would trigger an IsFunctioning() call.  It will now be checked for every 15 seconds, and if it starts responding, will be enabled and the scheduler kicked off.  This capability was created for the embedded platform (checking every second), but was never implemented in the standard server.

### DIFF
--- a/java/sage/NetworkCaptureDevice.java
+++ b/java/sage/NetworkCaptureDevice.java
@@ -344,7 +344,7 @@ public final class NetworkCaptureDevice extends CaptureDevice
   public boolean isFunctioning()
   {
     boolean rv = "OK".equals(submitHostCommand("NOOP"));
-    if (Sage.EMBEDDED && !rv && !launchedMonitorThread)
+    if (!rv && !launchedMonitorThread)
     {
       if (hasConfiguredInput())
       {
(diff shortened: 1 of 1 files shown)


TuscanRaider:
It is a good idea to make the duration a property instead of hardcoding it.

On Friday, November 27, 2015, Christopher Piper <notifications@github.com>
wrote:

> If a network encoder is configured and not found, it was previously
> ignored, unless the user navigated to the source details page, which would
> trigger an IsFunctioning() call. It will now be checked for every 15
> seconds, and if it starts responding, will be enabled and the scheduler
> kicked off. This capability was created for the embedded platform (checking
> every second), but was never implemented in the standard server.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/google/sagetv/pull/56
> Commit Summary
>
>    - Monitor lost network encoders
>
> File Changes
>
>    - *M* java/sage/NetworkCaptureDevice.java
>    <https://github.com/google/sagetv/pull/56/files#diff-0> (4)
>
> Patch Links:
>
>    - https://github.com/google/sagetv/pull/56.patch
>    - https://github.com/google/sagetv/pull/56.diff
>
> —
> <https://github.com/google/sagetv/pull/56>.
>


-- 
-Mike

CraziFuzzy:
That's possible - the 1000 msec duration was already in there hard coded on servers it was workign on (which actually I don't think there are any that are running v9, and embedded) - I just felt that was a little excessive, but didn't want to change the existing behavior.  While properties are great, you also want to avoid bloating up the .properties file with things that are unlikely to ever need to be changed by an end user.

PR #71, opened 2016-01-17
Added MENU_HINTS client property

stuckless:
whereby the client renderer on the server can send hints to the remote client letting it know about which menu or popup is active, and whether or not a keyboard is required.  Also added a new API IsTouchUI that returns TRUE if the remote client is a Touch enabled device.  The Server's Version is updated to 9.0.4 to reflect new APIs and Properties.

### DIFF
--- a/java/sage/MiniClientSageRenderer.java
+++ b/java/sage/MiniClientSageRenderer.java
@@ -1330,18 +1330,22 @@ public class MiniClientSageRenderer extends SageRenderer
 
   private boolean checkForMenuHintUpdates()
   {
-    if (menuHint.hasHint())
+    String hintText;
+    synchronized (menuHint) {
+      hintText=menuHint.format();
+      menuHint.clear();
+    }
+    if (hintText!=null)
     {
-      try
-      {
-        int propRV = sendSetProperty("MENU_HINT", menuHint.format());
+      try {
+        int propRV = sendSetProperty("MENU_HINT", hintText);
         if (propRV != 0)
         {
-          if (Sage.DBG) System.out.println("MiniClient did not succeed with menu hint change to:" + menuHint.menuName + ", errcode=" + propRV);
+          if (Sage.DBG)
+            System.out.println("MiniClient did not succeed with menu hint change to:" + hintText + ", errcode=" + propRV);
         }
-        menuHint.clear();
-      }
-      catch (java.io.IOException e)
+
+      } catch (java.io.IOException e)
       {
         if (Sage.DBG) System.out.println("Error w/ MiniClient UI:" + e);
         connectionError();
@@ -8013,9 +8017,11 @@ public class MiniClientSageRenderer extends SageRenderer
   @Override
   public void setMenuHint(String menuName, String popupName, boolean hasTextInput)
   {
-    menuHint.menuName=menuName;
-    menuHint.popupName=popupName;
-    menuHint.hasTextInput=hasTextInput;
+    synchronized (menuHint) {
+      menuHint.menuName = menuName;
+      menuHint.popupName = popupName;
+      menuHint.hasTextInput = hasTextInput;
+    }
   }
 
   private TextSurfaceCache getCachedTextSurfacePtr(RenderingOp op)
--- a/java/sage/MiniClientSageRenderer.java
+++ b/java/sage/MiniClientSageRenderer.java
@@ -1332,9 +1332,13 @@ public class MiniClientSageRenderer extends SageRenderer
   {
     String hintText;
     synchronized (menuHint) {
+      // menuHint.format() will return null if no hint is available
+      // otherwise get a formatted hint text to send to the client
       hintText=menuHint.format();
       menuHint.clear();
     }
+
+    // if we have a hint, then push it to the client
     if (hintText!=null)
     {
       try {
@@ -8827,13 +8831,20 @@ public class MiniClientSageRenderer extends SageRenderer
       hasTextInput=false;
     }
 
+    /**
+     * @return true if there is a menu or popup in this hint
+     */
     public boolean hasHint()
     {
       return menuName!=null || popupName!=null;
     }
 
+    /**
+     * @return formatted hint text, or null, if no hint is available
+     */
     public String format()
     {
+      if (!hasHint()) return null;
       return "menuName:" + menuName +
               ", popupName:" + popupName +
               ", hasTextInput:" + hasTextInput;
--- a/java/sage/MiniClientSageRenderer.java
+++ b/java/sage/MiniClientSageRenderer.java
@@ -1328,6 +1328,29 @@ public class MiniClientSageRenderer extends SageRenderer
     return true;
   }
 
+  private boolean checkForMenuHintUpdates()
+  {
+    if (menuHint.hasHint())
+    {
+      try
+      {
+        int propRV = sendSetProperty("MENU_HINT", menuHint.format());
+        if (propRV != 0)
+        {
+          if (Sage.DBG) System.out.println("MiniClient did not succeed with menu hint change to:" + menuHint.menuName + ", errcode=" + propRV);
+        }
+        menuHint.clear();
+      }
+      catch (java.io.IOException e)
+      {
+        if (Sage.DBG) System.out.println("Error w/ MiniClient UI:" + e);
+        connectionError();
+        return false;
+      }
+    }
+    return true;
+  }
+
   private boolean checkForDARUpdates()
   {
     if (lastSetDAR != vf.getDisplayAspectRatio())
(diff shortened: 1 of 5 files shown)


java/sage/MiniClientSageRenderer.java
```diff
@@ -1328,6 +1328,29 @@ private boolean checkForAudioOutputChangeUpdates()
     return true;
   }
 
+  private boolean checkForMenuHintUpdates()
+  {
+    if (menuHint.hasHint())
+    {
+      try
+      {
+        int propRV = sendSetProperty("MENU_HINT", menuHint.format());
```
Narflex: I'd suggest getting the format() first, and storing it in a local var and clearing the hint all in a block synchronized around the menuHint object. Then also in setMenuHint, synchronize the code in there around the menuHint object as well. I realize the other code isn't done this way...but it probably should have been. :)

java/sage/PseudoMenu.java
```diff
@@ -121,6 +117,9 @@ public void activate(boolean redo)
     }
 
     multipleTextInputs = comp.getNumTextInputs(0) > 1;
+
+    // notify the UI Renderer as to which Menu we are activating, so that it can pass it to the client as a HINT
+    uiMgr.getRootPanel().getRenderEngine().setMenuHint(widg.getName(), null, comp.getNumTextInputs(0)>0);
```
Narflex: This can be better optimized so that getNumTextInputs is only called once and the value is stored and then reused to set multipleTextInputs and also the hint properly.

Narflex:
Just a few minor changes I'm suggesting, I can merge it after those are taken care of.

stuckless:
Review that and see if matches what you had in mind.

java/sage/MiniClientSageRenderer.java
```diff
@@ -1328,6 +1328,33 @@ private boolean checkForAudioOutputChangeUpdates()
...
   }
 
+  private boolean checkForMenuHintUpdates()
+  {
+    String hintText;
+    synchronized (menuHint) {
+      hintText=menuHint.format();
+      menuHint.clear();
+    }
+    if (hintText!=null)
```
Narflex: You should still call hasHint() here...then get the hintText in the sync block after that. With this current implementation...hintText will never be null (since format doesn't look at the state) and you'll get the HINT property on every single command sequence sent over. :)

Narflex:
Just one issue still. The other change you updated is fine now.

java/sage/MiniClientSageRenderer.java
stuckless: That's a good point, but, calling hasHint outside of the syncronized block seems to defeat the purpose of actually synchronizing in the first place.  I think that format() should return null if there is no hint, and that way the entire acess to menuHint remains within a sync block.  Otherwise,  I have to put al the code if if(menuHint.hasHint()) inside of a sync block and my feeling was that would be holding the lock for too long.

stuckless:
Ok, so, I updated format() to return null when there isn't an hint, which, to be honest, I think is what it should have done in the first place, since, format() isn't a toString() it's basically a marshalling function, and in this case there isn't any data to marshall.

java/sage/MiniClientSageRenderer.java
Narflex: Calling hasHint() outside of the synchronized block should be perfectly
fine.  The state may get updated simultaneously with that check..but it
won't get cleared, so the return value from hasHint() is still valid
outside the sync block. I did want to avoid grabbing the lock every time
this checks.  The purpose of the synchronization was two-fold:

1. Ensure that all fields of MenuHint are from a single call to setMenuHint
when creating the format(). This is achieved by synchronizing the call to
format() as well as setMenuHint().
2. Ensure that when you clear() the MenuHint structure, that you are
clearing what you already took the format for (so you don't get the
format() from a prior setMenuHint(), then a new hint is set, and then you
clear it and never end up sending the new hint).  This is achieved by
having format() and clear() in the same sync block.

It's fine the way you have it now...it's not like one extra lock for each
rendering cycle is going to hurt anything...there's tons of locking already
that happens on each render cycle. :)  I'll approve it the way it is, I
just explained things above so you understood my line of thinking.  Of
course my solution suffers from only focusing on the singular usage that
currently is there...yours is more robust for other usages that may come in
later. :)

On Tue, Jan 19, 2016 at 3:56 PM, Sean Stuckless <notifications@github.com>
wrote:

> In java/sage/MiniClientSageRenderer.java
> <https://github.com/google/sagetv/pull/71#discussion_r50195941>:
>
> > @@ -1328,6 +1328,33 @@ private boolean checkForAudioOutputChangeUpdates()
> >      return true;
> >    }
> >
> > +  private boolean checkForMenuHintUpdates()
> > +  {
> > +    String hintText;
> > +    synchronized (menuHint) {
> > +      hintText=menuHint.format();
> > +      menuHint.clear();
> > +    }
> > +    if (hintText!=null)
>
> That's a good point, but, calling hasHint outside of the syncronized block
> seems to defeat the purpose of actually synchronizing in the first place. I
> think that format() should return null if there is no hint, and that way
> the entire acess to menuHint remains within a sync block. Otherwise, I have
> to put al the code if if(menuHint.hasHint()) inside of a sync block and my
> feeling was that would be holding the lock for too long.
>
> —
> <https://github.com/google/sagetv/pull/71/files#r50195941>.
>



-- 
Jeffrey Kardatzke
jkardatzke@google.com
Google, Inc.

stuckless: That makes a lot of sense.  I tend to be very lenient or very strict with
locking :)  In this case because it was "hints" I didn't add any locking,
since my feeling was that the chances of the player getting the wrong hint
would be minimal, and likely short lived (ie, next hint with a menu or
popup would rectify it).  But, your logic is pretty good and minimizes the
locking until it is needed.  If I get a chance I'll look at refactoring the
code based on your explanation.

Thanks,
Sean.

On Tue, 19 Jan 2016 at 19:28 Jeffrey Kardatzke <notifications@github.com>
wrote:

> In java/sage/MiniClientSageRenderer.java
> <https://github.com/google/sagetv/pull/71#discussion_r50198533>:
>
> > @@ -1328,6 +1328,33 @@ private boolean checkForAudioOutputChangeUpdates()
> >      return true;
> >    }
> >
> > +  private boolean checkForMenuHintUpdates()
> > +  {
> > +    String hintText;
> > +    synchronized (menuHint) {
> > +      hintText=menuHint.format();
> > +      menuHint.clear();
> > +    }
> > +    if (hintText!=null)
>
> Calling hasHint() outside of the synchronized block should be perfectly
> fine. The state may get updated simultaneously with that check..but it
> won't get cleared, so the return value from hasHint() is still valid
> outside the sync block. I did want to avoid grabbing the lock every time
> this checks. The purpose of the synchronization was two-fold: 1. Ensure
> that all fields of MenuHint are from a single call to setMenuHint when
> creating the format(). This is achieved by synchronizing the call to
> format() as well as setMenuHint(). 2. Ensure that when you clear() the
> MenuHint structure, that you are clearing what you already took the format
> for (so you don't get the format() from a prior setMenuHint(), then a new
> hint is set, and then you clear it and never end up sending the new hint).
> This is achieved by having format() and clear() in the same sync block.
> It's fine the way you have it now...it's not like one extra lock for each
> rendering cycle is going to hurt anything...there's tons of locking already
> that happens on each render cycle. :) I'll approve it the way it is, I just
> explained things above so you understood my line of thinking. Of course my
> solution suffers from only focusing on the singular usage that currently is
> there...yours is more robust for other usages that may come in later. :)
> … <#msg-f:1523842769132898788_>
> On Tue, Jan 19, 2016 at 3:56 PM, Sean Stuckless <notifications@github.com>
> wrote: In java/sage/MiniClientSageRenderer.java <#71 (comment)
> <https://github.com/google/sagetv/pull/71#discussion_r50195941>>: > @@
> -1328,6 +1328,33 @@ private boolean checkForAudioOutputChangeUpdates() >
> return true; > } > > + private boolean checkForMenuHintUpdates() > + { > +
> String hintText; > + synchronized (menuHint) { > +
> hintText=menuHint.format(); > + menuHint.clear(); > + } > + if
> (hintText!=null) That's a good point, but, calling hasHint outside of the
> syncronized block seems to defeat the purpose of actually synchronizing in
> the first place. I think that format() should return null if there is no
> hint, and that way the entire acess to menuHint remains within a sync
> block. Otherwise, I have to put al the code if if(menuHint.hasHint())
> inside of a sync block and my feeling was that would be holding the lock
> for too long. — Reply to this email directly or view it on GitHub <
> https://github.com/google/sagetv/pull/71/files#r50195941>.
> -- Jeffrey Kardatzke jkardatzke@google.com Google, Inc.
>
> —
> <https://github.com/google/sagetv/pull/71/files#r50198533>.
>
-- 

Sent from Android

PR #80, opened 2016-03-03
Fix smbmount problem in IOUtils.java

BobPhoenix:
Hope I did the Git commits and pull requests correctly.  As you will see in history I had the file in the wrong spot first but it should be fixed in the pull request submitted.  I hope this is what you had in mind for the change.


java/sage/IOUtils.java
```diff
@@ -1252,7 +1253,21 @@ public static int doSMBMount(String smbPath, String localPath)
           else
             smbOptions = "guest:";
         }
-        if (IOUtils.exec2(Sage.LINUX_OS ? new String[] { "smbmount", smbPath, localPath , "-o", smbOptions } :
+        // check to see if property exists if it doesn't check for smbmount with "which" command
+        if (IOUtils.has_smbmount == null)
+        {
+            String result = IOUtils.exec(new String[] {"which", "smbmount"});
+            // if nothing returned from "which" command then smbmount is not present so set property false
+            if (result == null || result.isEmpty())
```
Narflex: Can you change isEmpty() to use .length() instead? I'm pretty sure we don't have any Java 1.6 code in SageTV yet and isEmpty is a 1.6 API call.  (I'm not opposed to moving to Java 1.6, but I'd want it to be for a better reason)

java/sage/IOUtils.java
```diff
@@ -1252,7 +1253,21 @@ public static int doSMBMount(String smbPath, String localPath)
...
+            {
+                IOUtils.has_smbmount = Boolean.FALSE;
+            } else {
+                IOUtils.has_smbmount = Boolean.TRUE;
+            }
+        }
+        // set execution variable based on property value
+        String execSMBMount = IOUtils.has_smbmount.booleanValue() ? "smbmount" : "mount.cifs";
```
Narflex: The .booleanValue() is not needed, the compiler will do that automatically.

Narflex:
Just a couple little nits...fix those and I'll merge it. Thanks!

PR #82, opened 2016-03-09
add WinSDK 6.1 guids to guids.h

bryanzim:
define missing interfaces in qedit.h
fix missing resource.h file
remove strmiids.lib from linker input file

This pull request allows the building of the VS2015 project without the installation of the Windows 6.1 SDK which is deprecated.  I would like to modify the project to build 64 bit executables also but this was a first step.


JustFred51:
If SDK6.1 is no longer needed, please update "HOWTO-BuildWindowsNativeCode.txt" to remove references to the SDK and it's associated environment variable.

Narflex:
Also, please take a look at the CONTRIBUTING file. You need to agree to the Google CLA before I can accept code contributions for you (it's super easy to do).

bryanzim:
I will take a look at the CLA and also update the txt docuemnt and go through the projects to remove any last vestiges of the Windows 6.1 SDK environment variables. The problem with using an older SDK and building on a newer system is depending on where the old SDK is in the include chain it could bring in a lot more older headers than desired. 

I also noticed that there where a multiple of project types for selection.  Do we really need all those or could we boil down to just debug/release and future debug x64/release x64 and build everything?

PR #117, opened 2016-03-29
Fix crash when closing DVB recording on 64-bit systems

rflitcroft:


rflitcroft:
I signed it!

Narflex:
Thanks for the contribution!

Jeff Kardatzke

PR #119, opened 2016-03-31
Fix crash with Hauppauge HDPVR IR blaster in Linux 64 Bit

ycayer:


ycayer:
I signed it!

ycayer:
I Signed it!

Narflex:
Thanks for the contribution!

PR #136, opened 2016-05-11
Added ability to use a local tuning plugin with a network encoder

enternoescape:
…der.

### DIFF
--- a/java/sage/CaptureDevice.java
+++ b/java/sage/CaptureDevice.java
@@ -442,6 +442,25 @@ public abstract class CaptureDevice
     return rv;
   }
 
+  protected void doPluginTune(String tuneString) {
+    doPluginTune(tuneString, activeSource.getTuningPluginPort());
+  }
+
+  protected void doPluginTune(String tuneString, int plugPort)
+  {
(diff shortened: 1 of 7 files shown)


Narflex:
I haven't looked at the STV code yet...will do that tomorrow. For the Java code there's one error in the way it's done now (which is essentially an irrelevant error since HDHomeRunCaptureDevice runs on Mac or Linux, not Windows, and you never use a tuning plugin with that device in reality); and overall I'd recommend a slightly different approach.

This is how I would do it, as it has less code duplication and seems somewhat cleaner:

1. Only have one doPluginTune method in CaptureDevice and make it a non-static method with 2 parameters for CaptureDeviceInput and tuneString.
2. Override this method in IVTVCaptureDevice for the one case specific to it, and then have it call the super.doPluginTune method otherwise (that's why it's non-static in #1, so you can reference the fields in IVTVCaptureDevice when using it there).

This then avoids the 3 way choice in NetworkCaptureDevice, and fixes HDHomeRunCaptureDevice (since that's used on Mac/Linux) since the specifics of the platform type of tuning are handled within the method in CaptureDevice then.

Does that sound good to you?

enternoescape:
Sounds good.

enternoescape:
The changes online don't match the changes I actually made. I'm going to create a new pull.

PR #137, opened 2016-05-12
Added ability to use a local tuning plugin with a network encoder.

enternoescape:
This one looks correct. Something strange happened when I pushed my changes on the other branch.

### DIFF
--- a/java/sage/CaptureDevice.java
+++ b/java/sage/CaptureDevice.java
@@ -442,6 +442,25 @@ public abstract class CaptureDevice
     return rv;
   }
 
+  protected void doPluginTune(String tuneString) {
+    doPluginTune(tuneString, activeSource.getTuningPluginPort());
+  }
+
+  protected void doPluginTune(String tuneString, int plugPort)
+  {
+    if (!DirecTVSerialControl.DIRECTV_SERIAL_CONTROL.equals(activeSource.getTuningPlugin()))
+    {
+      SFIRTuner tunePlug = ExternalTuningManager.getIRTunerPlugin(activeSource.getTuningPlugin(),
+              plugPort);
+      if (tunePlug != null)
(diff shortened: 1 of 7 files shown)


Narflex:
Almost good...there's a little bit of 'trash' left over in the STV you edited that should be cleaned up. If you scroll all the way down to the bottom of it, you'll see the following listed right after the Global theme, all of which you can delete:

false
true
Don't Configure Tuning
Don't Configure Tuning (You may still....Encoder)

Just delete only these 4 widgets and then it'll be clean. :) You don't need to delete the menus that are under false & true, those references will disappear automatically once you delete false & true.

enternoescape:
I completely missed those. I wonder how I accidentally created them. I have a new problem. When I remove the extra entries (I tried this twice), it ends up adjusting the values of <Text Ref="62145" Name=""/> to <Text Ref="62142" Name=""/>. Basically, it's adjusting all of the Ref and ID attributes.

Narflex:
They were left dangling when you disconnected them from where you made the changes.

If it does other adjustments as well...that's likely fine. There's a diff tool in the Studio that I use to compare these to determine what really did change (Tools->STV UID File Difference)...diffing the XML files only sometimes yields usable results; so don't worry about what it's doing to the actual XML.

Narflex:
I can also just take the changes the way they are if you want and fix the STV myself...it'll only take me a minute to do it. :) Your call.

enternoescape:
It's ok. I'll commit the changes shortly.

Narflex:
Sorry, it didn't notify me about the changes you made here via email (apparently it doesn't do that when you just update the commit in the pull request); so I didn't see this until now. I'll take a look at the STV diff on Monday and then merge it then after I verify it's fine.

enternoescape:
No worries. I figured you were just being thorough.

PR #143, opened 2016-06-21
Enable media server remux

enternoescape:


java/sage/MediaServer.java
```diff
@@ -560,6 +594,10 @@ protected void sizeFile() throws java.io.IOException
           availSize = xcoder.getVirtualTranscodeSize();
           totalSize = xcoder.isTranscodeDone() ? availSize : getLargeFileSize(currFile.toString());
         }
+        if (remuxer != null)
+        {
+          availSize = totalSize = remuxer.getFileSize();
```
Narflex: You'll want to use getLargeFileSize(String) for the totalSize if the file is recording. This is used as an indicator in some of the playback clients to know that the file is currently recording (they do it by checking if totalSize > availSize); also be sure you handled the circular file case as that will be used for the preview window during playback (although your remuxer logic may already handle that case, I haven't got that far reviewing the code yet :) )

java/sage/MediaServer.java
```diff
@@ -1157,6 +1213,147 @@ else if (tempString.indexOf("XCODE_SETUP ") == 0 && !Sage.EMBEDDED)
...
+            }
+            else
+            {
+              if (Sage.DBG) System.out.println("MediaServer is writing in remux mode: " +
+                  tempString.substring(tempString.indexOf(" ") + 1));
+
+              StringTokenizer toker = new StringTokenizer(tempString.substring(12), " ");
+
+              String mode = toker.nextToken();
+              int outputFormat = toker.nextToken().equals("TS") ? MPEGParser2.REMUX_TS : MPEGParser2.REMUX_PS;
```
Narflex: You may want to use equalsIgnoreCase for these parameter tests...especially the one for TRUE since somebody may just end up printing out a boolean value into the text at some point.

java/sage/MediaServer.java
```diff
@@ -1157,6 +1213,147 @@ else if (tempString.indexOf("XCODE_SETUP ") == 0 && !Sage.EMBEDDED)
...
+              {
+                remuxer.disableInterAssist();
+                commBufWrite.clear();
+                commBufWrite.put(OK_BYTES).flip();
+              }
+              else if (config.equals("FORMAT"))
+              {
+                commBufWrite.clear();
+                commBufWrite.put((remuxer.isInitialized() ?
+                    remuxer.getContainerFormat().getFullPropertyString() + "\r\n" : "NULL\r\n").getBytes()).flip();
```
Narflex: Use ContainerFormat.getFullPropertyString(false). Otherwise it may include metadata which could potentially have \r\n chars in it (although in reality I don't think that could ever happen; but it's worth changing it just in case)

java/sage/MediaServerRemuxer.java
```diff
@@ -0,0 +1,1008 @@
...
+  private int data3;
+
+  /**
+   * Create a new media server remuxer with auto-detection.
+   * <p/>
+   * The input stream is assumed TS.
+   *
+   * @param fileChannel  The starting FileChannel to be used for writing.
+   * @param outputFormat The format of the remuxed stream.
+   * @param isTV         Is the incoming stream expected to have video?
```
Narflex: This should probably be commented as 'Is the incoming stream from a TV Tuner?' (since at the native level, that's what it is referring to)

java/sage/MediaServerRemuxer.java
```diff
@@ -0,0 +1,1008 @@
...
+   *
+   * @param filename The new filename to switch to.
+   * @param uploadId The upload ID to be used to authorize the new filename.
+   */
+  public void startSwitch(String filename, int uploadId)
+  {
+    synchronized (switchLock)
+    {
+      // Don't allow SWITCH to the same file.
+      if (!switching && new File(filename) != currentFile)
```
Narflex: The second comparison here will never be equal because you are constructing a new File object so it's object reference cannot be the same as an existing one. You should just compare the filenames as String values instead.

java/sage/MediaServerRemuxer.java
```diff
@@ -0,0 +1,1008 @@
...
+      if (Sage.DBG)
+        System.out.println("CANNOT perform fast switch on a non-H264/MPEG2 MPEG2 TS file!");
+      return 0;
+    }
+
+    if (outputFormat == MPEGParser2.REMUX_TS)
+    {
+      if (videoPid == -1)
+      {
+        if (Sage.DBG) System.out.println("CANNOT perform fast switch without a video PID!");
```
Narflex: Well...technically you can; you just don't care where you switch it at. :)

java/sage/media/format/MPEGParser2.java
```diff
@@ -0,0 +1,474 @@
...
+    CHANNEL,
+    PROGRAM,
+    XX_XX,
+    XX_XX_XX,
+    AUTO
+  }
+
+  static
+  {
+    if (System.getProperty("os.name").toLowerCase().contains("windows") ||
```
Narflex: You can use the WINDOWS_OS and LINUX_OS constants in Sage.java instead; it always loads first and those values are safe to use in other static initializers as well.

java/sage/media/format/MPEGParser2.java
```diff
@@ -0,0 +1,474 @@
...
+    {
+      inputStream = new java.io.FileInputStream(inputFile);
+      outputStream = new java.io.BufferedOutputStream(new java.io.FileOutputStream(outputFile));
+      remuxer = openRemuxer(REMUX_TS, outputFormat, isTV ? StreamFormat.ATSC : StreamFormat.FREE,
+          SubFormat.UNKNOWN, TuneStringType.CHANNEL, 0, 0, 0, 0, 0, outputStream);
+
+      if (remuxer == null)
+        return false;
+
+      byte[] buf = new byte[8084];
```
Narflex: I've seen a few of these odd large sizes in your code; you may want to include the match for why it is this size (when it's relating to 188, that's clear; or just some other random power of 2; but things like 8084 make me curious)

Narflex:
And be sure to reply on here after you make more changes; it doesn't send me emails about it unless you do (so I would forget to look at it again otherwise).

java/sage/MediaServer.java
enternoescape: Fixed.

java/sage/MediaServerRemuxer.java
enternoescape: Fixed.

java/sage/MediaServerRemuxer.java
enternoescape: I missed that one. Thanks. Fixed. That line is in there in case a network encoder is handling SWITCH and submits the command twice.

java/sage/MediaServerRemuxer.java
enternoescape: I reworded those. I borrowed some text from another Java class for the wording there. I've been trying to keep things reasonably consistent. :)

java/sage/media/format/MPEGParser2.java
enternoescape: Fixed. Thanks for the tip. I wasn't sure if there was a reason the old remuxer didn't use those values. I probably should have just asked.

java/sage/MediaServer.java
enternoescape: Fixed. There is a way for the network encoder to let the remuxer know that it needs to create a circular file.

java/sage/MediaServer.java
```diff
@@ -1157,6 +1213,147 @@ else if (tempString.indexOf("XCODE_SETUP ") == 0 && !Sage.EMBEDDED)
...
+              if (Sage.DBG) System.out.println("MediaServer is already in remux mode ignoring: " +
+                  tempString.substring(tempString.indexOf(" ") + 1));
+            }
+            else
+            {
+              if (Sage.DBG) System.out.println("MediaServer is writing in remux mode: " +
+                  tempString.substring(tempString.indexOf(" ") + 1));
+
+              StringTokenizer toker = new StringTokenizer(tempString.substring(12), " ");
+
```
enternoescape: Fixed. I was thinking about a lot of these error cases, but I decided not to address some of them because if you should not be using the wrong syntax and the media server connection closing is the worst side effect.

java/sage/MediaServer.java
enternoescape: Fixed. I see where you're coming from. I'm so used to everything always being caps for commands that it never crossed my mind.

java/sage/media/format/MPEGParser2.java
enternoescape: It's a multiple of 188. I can push a fairly large amount of data through to the remuxer at a time, but if it's not 188 byte aligned, then the remuxer provides alignment feedback. To avoid even needing that feedback, I have made the buffers multiples of 188. I'll spell that out better in the code by showing the math.

java/sage/MediaServerRemuxer.java
enternoescape: I'm not sure why this is, but when I used equals(), it thinks that
LogCabinLiving-CascadeCabinHuntinWashingtonsSnoqualmiePass-3209491-0.ts and
LogCabinLiving-GallatinCanyonCabinWithaRiverView-3209493-0.ts are exactly the same file. I don't think equals does what you would think it would do. I removed !switching temporarily to be certain that was the problem. Why I get that new objects usually will not be equal to each other, I don't understand how equals() would do this, yet it works fine on a HashMap.

enternoescape:
I think I have addressed the issues you pointed out. Other than the mistake I made comparing two files, none of them were significant enough to change anything in a noteworthy way, so I'm comfortable with this quick turnaround.

java/sage/MediaServerRemuxer.java
```diff
@@ -884,9 +887,9 @@ private void pushData(byte[] buffer, int offset, int length)
       if (buffer[offset] != 0x47 || !tsSynced)
       {
         int startingOffset = offset;
-        if (Sage.DBG) System.out.println("Remuxer is out of sync.");
+        if (Sage.DBG && tsSynced) System.out.println("Remuxer is out of sync.");
 
-        while (length - (offset + 377) > 0)
+        while (length - (offset + (188 * 2)) > 0)
```
Narflex: I'm fairly sure this needs to be 377, not 376 for the value. Because if length is 377 here and offset is zero, then the conditional is true. Then offset gets incremented to 1. Then it will index array element 377 (which I'm assuming could correspond to the length of that array in certain cases).

java/sage/MediaServerRemuxer.java
```diff
@@ -33,8 +33,10 @@
 public class MediaServerRemuxer
 {
   private static final int TS_ALIGN = 188;
-  private static final int MAX_TRANSFER = 33088;
-  private static final int MAX_INIT_BUFFER_SIZE = 20971400;
+  // The transfers need to be 188 byte aligned.
+  private static final int MAX_TRANSFER = 176 * TS_ALIGN;
```
Narflex: I do like how you show which values are multiple of 188 now; but there's still seemingly random values in here such as 176, 111550, 8388608 (which is 8MB, so just add a comment after it saying 8MB for clarity), 27888.  I'd be concerned changing them at a later time because I wouldn't know why those values were picked.  I'm not going to press you on it; so you can leave it this way if you want, but it's not a bad idea to be more clear about constant values like that if there's a reason behind them (and I quite often just put stuff like "random large buffer size", when I'm just picking some big number out of the air; then someone later would know there's no concern with changing it).

Narflex:
OK, just two more comments from me. One I think is an actual bug; the other is just commenting. I am leaving for vacation tomorrow; so if you can get the changes in today I can merge them before I leave.

java/sage/MediaServerRemuxer.java
enternoescape: You're correct. The offset could be 1 over 188 * 2 and that would cause an out of bounds exception. Due to the alignment buffering, the code mentioned here doesn't get hit very often, so it was easy to miss. While I'm not a Java veteran, small mistakes like that are quite embarrassing.

enternoescape:
I agree on the comments, it's always nice to see them when you're trying to figure out what the other guy was thinking. I always hate that when I'm looking at a project and comments are missing because the author thought the intent was obvious...and it's really not.

Narflex:
Yeah! New remuxer is merged. Thanks for all the hard work!

PR #148, opened 2016-06-24
Allow Ministry (Transcoding queue) to keep Server awake.

CraziFuzzy:
Added a function requiresPower() to Ministry, and added a check to ServerPowerManagement to check it.  This functionality clones how ServerPowerManagement checks Seeker for activity.

### DIFF
--- a/java/sage/Ministry.java
+++ b/java/sage/Ministry.java
@@ -825,6 +825,15 @@ public class Ministry implements Runnable
       return true;
   }
 
+  public boolean requiresPower()
+  {
+    // This is true if there's any transcode jobs are in the queue
+    synchronized (this)
+    {
+    	return (getTranscodeJobIDs().length > 0) ? true : false ;
+    }
+  }
+  
   private Object lock = new Object();
   private java.util.Vector waitingForConversion = new java.util.Vector();
   private java.util.Vector converting = new java.util.Vector();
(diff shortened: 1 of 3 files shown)


java/sage/NetworkEncoderManager.java
```diff
@@ -64,7 +64,7 @@ public void detectCaptureDevices(CaptureDevice[] alreadyDetectedDevices)
       }
     }
 
-    if (Sage.getBoolean("network_encoder_discovery", false))
+    if (Sage.getBoolean("network_encoder_discovery", true))
```
Narflex: Can you just make this another commit please? I'm fine with this change; but I don't want it mixed into a commit for something totally unrelated.

java/sage/Ministry.java
```diff
@@ -825,6 +825,15 @@ public boolean isValidSourceFormat(sage.media.format.ContainerFormat cf)
       return true;
   }
 
+  public boolean requiresPower()
+  {
+    // This is true if there's any transcode jobs are in the queue
+    synchronized (this)
+    {
+    	return (getTranscodeJobIDs().length > 0) ? true : false ;
```
Narflex: You could just make this "return (getTranscodeJobIDs().length > 0);"

java/sage/NetworkEncoderManager.java
Narflex: Oh wait, nevermind...this is actually a separate commit; it was just grouped into one pull request for me. :)

PR #151, opened 2016-06-26
The transition (SWITCH) point for MPEG2-PS would sometimes never be found because of signed bytes in Java.

enternoescape:
I missed this one since my test cases actually passed without this modification, but just the other day I caught a few transition that timed out because over 8MB of data was processed without finding a transition point. The root of the problem is signed bytes. This masks the bytes so we are comparing the correct "unsigned" value.

### DIFF
--- a/java/sage/MediaServerRemuxer.java
+++ b/java/sage/MediaServerRemuxer.java
@@ -604,7 +604,7 @@ public class MediaServerRemuxer
         if (data.get(i) == 0x00 &&
             data.get(i + 1) == 0x00 &&
             data.get(i + 2) == 0x01 &&
-            data.get(i + 3) == 0xBA)
+            (data.get(i + 3) & 0xFF) == 0xBA)
         {
           psStart = i;
 


CraziFuzzy:
Could this looped search be done more efficiently with data.getLong(i) and making a single comparison?  I'm not sure about the performance in java of all these single byte data moves and comparisons.

Even faster MIGHT be to use a Long variable to load in at first, and then shift 8 bits and add a data.get(i) byte as it loops through.

Honestly, I'm not even sure how often this search is being done - just my quick observation at seeing this small bit of code.

enternoescape:
This is only used to find a good transition point to start a new file. This particular search is really not my code; it's adapted from what is done in the C++ code and is only used for about half a second typically, so I think it's good enough, especially since I have yet to see this code spike the CPU. Think about how slowly the video actually comes in.

This is technically more efficient than getLong because instead of bit shifting 4 times just to find out the first byte isn't the right value, we find out rather quickly. I could have it do the kind of bit shifting that's used to find a sequence header once we reach a valid pack header, but I really don't think this merits that level of optimization given how little it's actually "combing" over the stream.

CraziFuzzy:
Makes sense, like i said, It was just an observation on a small bit of the code.  I don't know much about Java performance, and thought that dealing with a single 64-bit number might be easier (especially if more and more run a 64-bit JVM going forward).

enternoescape:
I really do appreciate more eyes on what's going into the core. Also I just realized, that would be bit shifting 8 times for long; that's even worse since we only need the first 4 bytes. :) I got hung up on 4 bytes because that's how many we are actually using. I really do consider these things especially when there's a convenience method like that, but often it doesn't actually fit what I need or would do more than I need, so I don't bother with it.

In Java all numbers are signed including bytes. It really makes the language less suitable for processing A/V because it adds a different type of error you can accidentally introduce. Basically the code you see here was originally comparing a byte value to an integer value. 0xBA is actually a negative byte. When compared, Java will keep the sign. So the comparison ends up being (int) 186 != (byte) -70.

Narflex:
Oh yeah...this fun Java issue. :)  I get so tired of ANDing byte values with 0xFF just to compare them against a hex constant in Java....it's all throughout the SageTV codebase.

PR #156, opened 2016-07-18
Fixed issue #87 so that now SRT subtitles are read live and made local and remote file access consistent and easily extendable.

enternoescape:
This does not address multi-part recordings that have a subtitle file for each segment.


java/sage/Wizard.java
```diff
@@ -15,66 +15,34 @@
...
-
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
-import org.apache.lucene.index.CorruptIndexException;
-import org.apache.lucene.index.IndexCommit;
-import org.apache.lucene.index.IndexReader;
-import org.apache.lucene.index.IndexWriter;
-import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.*;
```
Narflex: My personal preference is to never import .*. I've been bitten by that before where it was using the wrong class because I wasn't aware something with the same class name was in a .* imported package.

enternoescape: I can change that back, I left cleanup checked when I committed and it consolidated those automatically. I've noticed your preference to not import.

Narflex: OK, it changed my comment when I put a "dot star" and apparently inferred that to mean italics...odd. I'm totally fine with imports, its just the wildcard ones I'm not a fan of.

Narflex:
Can you provide a summary of the new file/IO classes you've created? This is a huge change, so it's going to take me some time to go over all of it and any additional information (aside from what we've already talked about in other comments) will help speed that up. :)  It'll also help anybody else trying to understand how to use all of this.

enternoescape:
These classes all use IO:
sage.io.SageFileSource - An implementation used by the classes sage.io.LocalSageFile, sage.io.RemoteSageFile, sage.io.BufferedSageFile and sage.io.EncryptedSageFile. It provides a standardized framework for IO operations in SageTV. The idea is that each one tries to focus on doing one thing well so if for example we come up with a 3rd way we want to read a file/data or maybe even an ISO, you only need to write exactly that part and don't need to think about things like buffering or adding methods to get a single byte, short int, long. They are designed to be stackable with an actual data source at the bottom.

sage.io.LocalSageFile - This is 95% RandomAccessFile.

sage.io.RemoteSageFile - This uses the media server to access files.

sage.io.BufferedSageFile - This adds a read and write buffer layer on top of any class that implements sage.io.SageFileSource. It does 100% buffered writing and it tries to determine based on the sizes of reads if it needs to buffer fully, some combination or fully read directly from it's source.

sage.io.EncryptedSageFile - This adds an encryption and decryption layer on top of any class that implements sage.io.SageFileSource. This is the same encryption as the sage.FastRandomAccess file class.

These classes add other data processing capabilities over top of the sage.io.SageFileSource implementations:
sage.io.SageDataFile - Adds java.io.DataInput and java.io.DataOutput implementations over top of any sage.io.SageFileSource implementation.

sage.io.SageInputStream - Converts any sage.io.SageFileSource into an java.io.InputStream.

sage.io.SageOutputStream - Converts any sage.io.SageFileSource into an java.io.OutputStream.

sage.io.SageIOFileChannel - Implements FileChannel over any sage.io.SageFileSource. This one is dependent on sage.io.BufferedSageFile. If is it not present at the top level, it will automatically wrap it around the provided class, but it will then always flush the read or write buffers after every write to ensure you get the expected behavior. This of course negates any performance benefits. It needs the buffer layer since it already has byte arrays allocated to transfer between ByteBuffer and byte[].

These classes all use NIO:
sage.nio.SageFileChannel - This interface implements most of what you would expect to see in a FileChannel, plus it allows you to change out the source file which was needed for the transitioning between files in sage.media.bluray.BlurayFile. This interface is meant to be stackable similar to sage.io.SageFileSource. It has the methods transferTo(long, WritableByteChannel) and transferFrom(ReadableByteChannel, long) which transfer data relative to the current position in the file. tryLock(), lock(), map() and their associated methods are non-functional partly because we don't have any great use for them in SageTV; this is true for all of the classes extending FileChannel that I wrote.

sage.nio.LocalFileChannel - This is nearly a direct implementation of FileChannel and also extends FileChannel with some extra methods per the sage.nio.SageFileChannel interface.

sage.nio.RemoteFileChannel - This accesses files via the media server. It also extends FileChannel with some extra methods per the sage.nio.SageFileChannel interface.

sage.nio.BufferedFileChannel - This wraps around any sage.nio.SageFileChannel and provides both read and write buffer. Depending on how assumed large the transfer per the read or write method determines just how much read or write buffering there really is. For example, the scatter and gather methods will use what's left in the read/write buffer, then read directly since it's expected that you must be doing very large reads to need to use those methods.

sage.media.bluray.BlurayFile - This replaces sage.media.bluray.BlurayNetworkFile and sage.media.bluray.BlurayRandomFile and implements BlurayStreamer. It also extends FileChannel, though that was probably a little much in that case. We now only have one class to maintain for the same functionality. I left all of the old classes because these are somewhat big changes. It wraps sage.nio.BufferedFileChannel around sage.nio.LocalFileChannel or sage.nio.RemoteFileChannel depending on if we have local access or not.

java/sage/Wizard.java
enternoescape: The comments are in markdown, so you need to escape certain characters. \* It bugs me too sometimes.

I do strive to be a good contributor to this open source project and I have tried to emulate some of the style I've seen even when it's not my preference. I will fix the imports.

Narflex:
Can you increment the version number? This is a significant enough change with wide reach that I think it warrants a new microversion increment. :)

java/sage/io/SageFileSource.java
```diff
@@ -0,0 +1,277 @@
...
+   * @throws IOException If there is an I/O related error.
+   */
+  public void close() throws IOException;
+
+  /**
+   * Is this file actively growing?
+   *
+   * @return <code>true</code> if the file is actively growing.
+   */
+  public boolean isActiveFile();
```
Narflex: I don't think you should have the isActiveFile() method in here. That's a much higher level concept, and it's also not actually implemented except for network files so I don't want people thinking they can use it. There is MMC.isRecording(File) which will give you the correct answer...BUT that does not work properly unless you are in the server process (i.e. there's no way to call this on SageTVClient, but yes, the client should be accessing files remotely, but for the service case there may be optimizations where it is not).

As I also don't see any actual usages of this yet in the code, removing it doesn't seem like a problem.

java/sage/io/SageFileSource.java
```diff
@@ -0,0 +1,277 @@
...
+
+  /**
+   * Execute a command specific to this implementation.
+   *
+   * @param command The command to execute.
+   * @return The result from the command. This could be null. The response is implementation
+   *         specific.
+   * @throws IOException If there is an I/O related error.
+   */
+  public String executeCommand(String command) throws IOException;
```
Narflex: The executeCommand methods seem very odd for a file class. :) It looks like they are just being used to send commands for network files...which makes sense; so in that case, I'd think you would just make them part of the RemoteFileChannel class (and currently they are only used internally within that class, so they wouldn't even need to be public).

I do understand that maybe you put it at this level so then they could push through all the wrappers properly...but as all the non-network cases would just ignore the command, it seems like usage of it would be ambiguous unless you actually knew you were dealing with a RemoteSageFile.  So if that's what you were thinking, maybe then when you need this access later you just create a static method somewhere which can get the RemoteSageFile source from an arbitrary SageFileSource object.

Feel free to explain further, I may be totally missing the point of this. :)

enternoescape: I think you hit that one on the head. As I got further into things, I realized that there may not be any benefit. I left it in there because maybe we would find some special case where it would be helpful, but you saying it's odd is enough that I'll just remove it from the interface.

java/sage/io/LocalSageFile.java
```diff
@@ -0,0 +1,235 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sage.io;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+
+public class LocalSageFile extends RandomAccessFile implements SageFileSource
```
Narflex: I'm curious why you chose to extend RandomAccessFile rather than make it a member inside of here? IIRC, the reason I didn't extend it was because I didn't want to end up using methods that I didn't expect to be used. I notice in yours you make some methods not supported, and so generally you're changing this to really not be like java.io.RandomAccessFile anymore (which is fine); but in that case you probably shouldn't make it extend it.

This class is also more about general byte access into a local file...and RandomAccessFile exposes all those other DataInput/DataOutput methods that it also seems to go beyond the scope of what LocalSageFile was intended to cover.

enternoescape: I didn't extend it at first, but then I did in what could be a misguided attempt to have one less method between the top class and the actual data source. I can see how someone might accidentally use that incorrectly. I'll make it less exposed.

java/sage/io/SageFileSource.java
enternoescape: This makes sense. I removed it.

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,489 @@
...
+    if (transcodeMode != null && transcodeMode.length() > 0)
+    {
+      outStream.write(("XCODE_SETUP " + transcodeMode + "\r\n").getBytes(Sage.BYTE_CHARSET));
+      outStream.flush();
+      String response = Sage.readLineBytes(inStream);
+      if (!"OK".equals(response))
+        throw new java.io.IOException("Error with remote transcode setup for " + transcodeMode + " of: " + response);
+    }
+
+    outStream.write("OPENW ".getBytes(Sage.BYTE_CHARSET));
```
Narflex: I was curious what you were going to do with the 'readonly' field here. :)  When you open a remote file with the media server...you can open it either in read mode or write mode; but not both at the same time (although the MediaServer could be extended to support such a case easily by opening a RandomAccessFile and getting the channel from there rather than a FileOuptutStream object in the openWriteFile method).

The way this is currently done; any calls to the write method in here will fail because the files are being opened in read mode on the server always.  Since you've got all the other work in here...I'd say just change MediaServer to use a RandomAccessFile's channel instead of FileOutputStream (the uploadStream field); then make this use OPENWRITEW instead of OPENW if readonly is set to false (and prior to that, request the upload key from the server so it'll grant you write access). That *should* have no performance implications.

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+  @Override
+  public synchronized long skip(long n) throws IOException
+  {
+    if (n <= 0)
+      return 0;
+
+    long pos = remoteOffset;
+    long seek = Math.min(pos + n, length());
```
Narflex: You should probably optimize it to avoid calling length() unless required. You have this check in a few places, so you may just want to make a private method that does it to prevent code duplication (i.e. isLengthGreater(long))

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,489 @@
...
+  @Override
+  public long position()
+  {
+    return remoteOffset;
+  }
+
+  @Override
+  public long length() throws IOException
+  {
+    String response = executeCommand("SIZE\r\n");
```
Narflex: You can optimize this so that if the first time you send SIZE, you get back the same value for both, you never need to call the remote command again. Once a file goes non-active...it will never go active again.

java/sage/io/EncryptedSageFile.java
```diff
@@ -0,0 +1,327 @@
...
+   *
+   * @param sageFileSource The <code>SageFileSource</code> to wrap inside this filter.
+   * @param crypt Specify alternative 128 byte key.
+   * @throws IOException If there is an I/O related error or the <code>SageFileSource</code> is
+   *                     another instance of <code>EncryptedSageFile</code>.
+   */
+  public EncryptedSageFile(SageFileSource sageFileSource, byte crypt[]) throws IOException
+  {
+    // In case we try to make a file that's already using this "filter" use it twice in a row.
+    if (sageFileSource instanceof EncryptedSageFile)
```
Narflex: I wouldn't say this is an exception...of course there's no need for it in the current code...but maybe somebody wants to double encrypt in the future for some reason. :)  I'm fine with it either way though.

Narflex: Ahhh...OK, but this could then break getting the unencrypted source back properly, so leave it in here...BUT, you should probably loop all the way down the chain to make sure another EncryptedSageFile isn't hiding somewhere else in there. (i.e. keep getting the source of the source until it is null or you find another encrypted one and throw an exception).

java/sage/io/RemoteSageFile.java
enternoescape: That's only partly true. When the SRT file is growing, the media server as far as I've been able to tell provide any solid indication that is in fact happening.

java/sage/io/EncryptedSageFile.java
```diff
@@ -0,0 +1,309 @@
...
+   * length are preserved within the array.
+   *
+   * @param b The byte array to be encrypted.
+   * @param off The offset of the data to be encrypted.
+   * @param len The number of bytes from the offset to encrypt.
+   * @param fp The position of these bytes relative to the start of the file.
+   * @return A shared byte array containing the now encrypted bytes at the same offset as the
+   *         provided array.
+   */
+  protected byte[] encryptWrite(byte[] b, int off, int len, long fp)
```
Narflex: I'd make this method synchronized because of writeBuffer access.

java/sage/io/RemoteSageFile.java
enternoescape: I believe I mis-followed what you were doing to get the FileChannel in the media server. There's also a FileChannel.open() static method which is what I would probably use since that will result in just getting a file channel without creating any other classes we aren't actually going to use.

java/sage/io/BufferedSageFile.java
```diff
@@ -0,0 +1,880 @@
...
+  }
+
+  @Override
+  public int read(byte[] b) throws IOException
+  {
+    return read(b, 0, b.length);
+  }
+
+  @Override
+  public int read(byte[] b, int off, int len) throws IOException
```
Narflex: So I know this isn't readFully...BUT, I'm a little worried that this may bring out other bizarre bugs buried inside of SageTV because this will very often not read the requested amount if the buffer has any data left in it; whereas the implementations it is replacing would end up doing that (based on that other discussion we had). I'd feel safer if this actually read as much as it could before returning.

java/sage/io/RemoteSageFile.java
enternoescape: I re-optimized the file length() method per your suggestion and it broke the live subtitles. It's clear that it uses length() to see if there's more data.

java/sage/io/BufferedSageFile.java
```diff
@@ -0,0 +1,880 @@
...
+  private byte writeBuffer[];
+  // The write offset is always ahead of the actual file offset by writeOffset.
+  private int writeOffset = 0;
+  // This is used for random writes and seeking to indicate that we are actually ahead or the
+  // current write position for commits. If this value is not 0, we use this instead of writeOffset
+  // to determine how much to write.
+  private int writeLength = 0;
+  // This value is used to determine how many bytes to leave in the write buffer to optimize random
+  // writing. This value is adjusted dynamically based on how far behind the last random write was.
+  private int writeOptimizer = 0;
```
Narflex: Can you explain more what the whole write optimization thing is about? It's somewhat confusing, especially around the usage of writeLength and what it means....but mainly, why it's being done in the first place. I'm assuming it has something to do with an analysis you did regarding how Wizard is writing out files where it always skips backwards to write command lengths and this was done to reduce the number of overall actual disk writes.

Narflex:
Note to self...files reviewed:
build.gradle
SageFileSource (but I had comments on it regarding isActiveFile and executeCommand)
LocalSageFile (although I have issues about extending RandomAccessFile rather than wrapping it)
RemoteSageFile (with comments about write mode)
EncryptedSageFile (minor comment)
BufferedSageFile (in progress...I'm confused about writeLength currently)

java/sage/io/RemoteSageFile.java
Narflex: Ohhh...right, you're looking at files generated by some other process...not SageTV, so it has no knowledge of whether they are growing or not. Hmmm...that makes it more tricky...and that problem was the main motivation for this work. :) You can put it back the way it was then...especially considering the current code didn't have the optimization I suggested either.

java/sage/io/RemoteSageFile.java
Narflex: If you change that behavior; just make sure you're opening it in the proper read/write mode since that's how read/write security is enforced in MediaServer.

java/sage/io/RemoteSageFile.java
enternoescape: I did what I think is a good compromise. I added another constructor that enables you to force it to still check.

java/sage/io/BufferedSageFile.java
enternoescape: I know the Wizard only goes backwards when it random writes. I accounted for both directions in case that's actually useful at some point. writeLength is there in case we seek backwards inside of the writeBuffer or we write just in front of writeOffset. It ensures that everything eventually gets written.

The write optimization thing is triggered is a write goes behind the the last flushing of the buffer. The value is set based on how far back the write went unless it was more than half of the write buffer, then the value is decremented since it's proving to be less useful and at that point full writes are more valuable. The value basically indicates how much to leave in the write buffer.

java/sage/io/RemoteSageFile.java
enternoescape: I removed FileOutputStream and replaced it with FileChannel.open(currFile.toPath(), StandardOpenOption.CREATE, StandardOpenOption.READ, StandardOpenOption.WRITE)

java/sage/io/RemoteSageFile.java
enternoescape: Noted and addressed.

java/sage/io/RemoteSageFile.java
enternoescape: It calls read(byte[],int,int) and it was returning 0 which I fixed to return -1 which is what should be happening per the contract. I still added an initialization parameter to force the file to be thought active so that SRT subtitles can be read live.

java/sage/io/EncryptedSageFile.java
enternoescape: I suppose the way this is built that's a real possibility that one could be buried in the stack. This is here to help out SageDataFile so it doesn't need to be aware of multiple encryption layers. Either way, I made the search recursive per your suggestion.

java/sage/io/BufferedSageFile.java
enternoescape: I fixed this to just call readFully with the length being the minimum of what's remaining or the length itself. It's worth noting that based on my usage searches, I couldn't find a single class that actually used read(byte[],int,int). It looks like everything is using readFully which really is the more efficient choice if you know your file size.

java/sage/io/BufferedSageFile.java
enternoescape: It tries to effectively get the back writing re-aligned with the buffer by retaining some of the write instead of committing it with the expectation that the data we would have written out, then almost immediately modified on disk instead of in memory. I set up some logging to verify that it was actually working and on even a small database, it was able to help get us re-aligned over 40 times. Also just increasing the write buffer from 64k to 128k made a big difference in the number of times this optimization kicked in.

enternoescape:
I incremented the version in Version.java. Is there anywhere else I should be updating?

enternoescape:
I found the thread below and updated the native versions too. I think I have them all now.

http://forums.sagetv.com/forums/showthread.php?t=63296

java/sage/io/BufferedSageFile.java
```diff
@@ -0,0 +1,865 @@
...
+          if (bytesWrite == -1)
+            break;
+
+          writeOffset += bytesWrite;
+          remaining -= bytesWrite;
+
+          // You must set writePending to true before seeking or the data will be appended to or
+          // overwrite the wrong place in the file.
+          if (writeOffset == writeBuffer.length)
+            flushForced();
```
Narflex: Curious why you are using flushForced() here instead of flushOptimized()

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,555 @@
...
+      outStream.flush();
+      String response = Sage.readLineBytes(inStream);
+      if (!"OK".equals(response))
+        throw new java.io.IOException("Error with remote transcode setup for " + transcodeMode + " of: " + response);
+    }
+
+    if (readonly)
+      outStream.write("OPENW ".getBytes(Sage.BYTE_CHARSET));
+    else
+      outStream.write("WRITEOPENW ".getBytes(Sage.BYTE_CHARSET));
```
Narflex: WRITEOPENW takes 2 paramters, the first is the file path, and the second is the uploadID. That's an integer value that's retrieved from the server to prevent any other network service form arbitrarily opening a file and being able to write to it. I'm fine with you making that happen as part of this automatically. The way you get that on the client is calling SageTVConnection.requestUploadSpace(File, long) which specifies a file path and how big the upload is going to be. It uses the file size for freeing up space if needed in case that file path is in a controlled directory. However, apparently that takes zero size to clear the request...so you need to put some size in there. It also fails if the file already exists. (look at SageTVConnection.recvRequestUploadSpace which is the server end of the implementation). Maybe make a -1 file size a special case in there which doesn't free up any space, disregards checks for existing files and just gets a new upload ID.

Narflex: And if you do the uploadID thing, be sure to also release it when the file is closed (just request it again with zero size, that clears it on the server).

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,555 @@
...
+    // We can't read more than what is in the file. The first part checks a cached variable, the
+    // second part checks the file directly. Unless the stream really lingers trying to get a
+    // massive read towards the end, this should be a very minimal load on the media server.
+    count = Math.min(
+        maxRemoteSize == 0 || (position + count >= maxRemoteSize && activeFile) ?
+            reallyGetLength() - position : maxRemoteSize - position,
+        count);
+
+    // If we are trying to read beyond the end of the file, don't read anything.
+    return (count == 0) ? 0 : count;
```
Narflex: Why not just return count? :)

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,555 @@
...
+      outStream.write(("WRITE " + remoteOffset + " " + len + "\r\n").getBytes(Sage.BYTE_CHARSET));
+      outStream.flush();
+      outStream.write(b, off, len);
+    }
+
+    remoteOffset += len;
+  }
+
+  @Override
+  public void randomWrite(long pos, byte[] b, int off, int len) throws IOException
```
Narflex: This should be synchronized since you are changing remoteOffset in this method which can be changed in our synchronized methods. (of course you're totally asking for trouble if you're using a file stream that way anyways...but hey...correctness and all :) )

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,555 @@
...
+  public synchronized void write(int b) throws IOException
+  {
+    if (readonly)
+      throw new IOException("Remote file is read only.");
+
+    try
+    {
+      outStream.write(("WRITE " + remoteOffset + " 1\r\n").getBytes(Sage.BYTE_CHARSET));
+      outStream.flush();
+      outStream.write(b);
```
Narflex: Flush after you write the data....same thing for all the places you write data.

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,555 @@
...
+   * If \r\n is missing, it will automatically be appended for you. Everything sent with this method
+   * will be encoded as Sage.BYTE_CHARSET.
+   *
+   * @param command The command to execute.
+   * @return The response received from the media server.
+   * @throws IOException If there was an I/O error.
+   */
+  public String executeCommand(String command) throws IOException
+  {
+    command = command.endsWith("\r\n") ? command : command + "\r\n";
```
Narflex: Put parenthesis around (command + "\r\n") for clarity.

java/sage/io/SageIOFileChannel.java
```diff
@@ -0,0 +1,252 @@
...
+      this.randomFileSource = new BufferedSageFile(sageFileSource, BufferedSageFile.READ_BUFFER_NIO_LIMIT, BufferedSageFile.WRITE_BUFFER_SIZE);
+      alwaysFlush = true;
+    }
+  }
+
+  @Override
+  public int read(ByteBuffer dst) throws IOException
+  {
+    int returnValue = randomFileSource.read(dst);
+    if (alwaysFlush)
```
Narflex: Do we really need to flush on a read?

java/sage/nio/RemoteFileChannel.java
```diff
@@ -0,0 +1,703 @@
...
+  private void open() throws IOException
+  {
+    if (remoteFilename == null)
+      return;
+
+    commBuf.clear();
+    if (readonly)
+      commBuf.put("OPENW ".getBytes(Sage.BYTE_CHARSET));
+    else
+      commBuf.put("WRITEOPENW ".getBytes(Sage.BYTE_CHARSET));
```
Narflex: Same issue with usage of WRITEOPENW as in RemoteSageFile class.

java/sage/nio/RemoteFileChannel.java
```diff
@@ -0,0 +1,703 @@
...
+  {
+    // We can't read more than what is in the file. The first part checks a cached variable, the
+    // second part checks the file directly.
+    count = Math.min(
+        maxRemoteSize == 0 || (position + count >= maxRemoteSize && activeFile) ?
+            reallyGetSize() - position : maxRemoteSize - position,
+        count);
+
+    // If we are trying to read beyond the end of the file, don't read anything.
+    return (count == 0) ? 0 : count;
```
Narflex: Can you just return count?

java/sage/nio/RemoteFileChannel.java
```diff
@@ -0,0 +1,703 @@
...
+      // Since we are iterating through the buffers either way, we might as well find the first one
+      // that actually has space available.
+      if (remains > 0)
+      {
+        bufferAvailable = true;
+        totalLength += remains;
+      }
+      else if (!bufferAvailable)
+      {
+        offset += 1;
```
Narflex: If a buffer partway into the array has no room, then this may skip an earlier buffer. Such as we have A, B, C with zero offset and B is already full. Ater this, offset will be 1 and will just use B and C and will have skipped A which has room.

java/sage/nio/RemoteFileChannel.java
```diff
@@ -0,0 +1,703 @@
...
+      // Since we are iterating through the buffers either way, find the first one that actually has
+      // data available.
+      if (remains > 0)
+      {
+        bufferAvailable = true;
+        totalLength += remains;
+      }
+      else if (!bufferAvailable)
+      {
+        offset += 1;
```
Narflex: Same issue with offset as in the prior comment.

java/sage/io/BufferedSageFile.java
enternoescape: You're right, I'm not sure why I did that. Fixed it.

java/sage/io/RemoteSageFile.java
enternoescape: I added the uploadID as a parameter in the constructor so there's more transparency to the person using this about what's really happening and fixed the missing parameter.

java/sage/io/RemoteSageFile.java
enternoescape: That was changed a few times and the end result is like that because I didn't notice I had changed it into something obvious.

java/sage/io/RemoteSageFile.java
enternoescape: I agree. Done.

java/sage/io/RemoteSageFile.java
enternoescape: That makes sense.

java/sage/nio/RemoteFileChannel.java
enternoescape: Fixed it there too similar to how I fixed it in RemoteSageFile.

java/sage/nio/RemoteFileChannel.java
enternoescape: Fixed that.

java/sage/nio/BufferedFileChannel.java
```diff
@@ -0,0 +1,967 @@
...
+  {
+    if (readonly)
+      throw new NonWritableChannelException();
+
+    if (readBuffer.limit() != 0)
+      clearReadBuffer();
+    else if (writePending)
+    {
+      if (writeBuffer.position() < writeLength)
+        writeLength = writeBuffer.position();
```
Narflex: I'm not sure on this one...it seems by shortening writeLength here to match writeBuffer.position() that you are invalidating anything in the writeBuffer that's past the position but before the length. Is the comparator inverted here?

java/sage/nio/RemoteFileChannel.java
enternoescape: I thing I see what you're saying.

java/sage/media/bluray/BluRayFile.java
```diff
@@ -0,0 +1,496 @@
...
+      sageFileChannel = new BufferedFileChannel(new RemoteFileChannel(hostname, (String)null, true), readBufferSize, directBuffer);
+
+    this.bdmvDir = bdmvDir;
+    int targetTitleTmp = targetTitle;
+    bdp = new sage.media.bluray.BluRayParser(bdmvDir, hostname);
+    bdp.fullyAnalyze();
+    if (targetTitleTmp <= 0)
+      targetTitleTmp = bdp.getMainPlaylistIndex() + 1;
+    this.targetTitle = Math.max(1, Math.min(targetTitleTmp, bdp.getNumPlaylists()));
+    currPlaylist = bdp.getPlaylist(targetTitleTmp - 1);
```
Narflex: To match the old code, this should change (targetTitleTmp - 1) to be (this.targetTitle - 1). May be the same...but rather than digging through it all to analyze, just change this var name. :)

java/sage/media/bluray/BluRayFile.java
```diff
@@ -0,0 +1,496 @@
...
+  {
+    return sageFileChannel.position() + fileOffsets[currFileIndex];
+  }
+
+  @Override
+  public long skip(long n) throws IOException
+  {
+    long seek = Math.min(position() + n, totalSize);
+    ensureProperFile(false, seek);
+    return sageFileChannel.skip(n - fileOffsets[currFileIndex]);
```
Narflex: This isn't right because if it doesn't open a new file, it'll skip byte from the file start rather than the offset. To simplify this, I'd say just call position(seek).

java/sage/media/bluray/BluRayFile.java
```diff
@@ -0,0 +1,496 @@
...
+  public long transferTo(long position, long count, WritableByteChannel target) throws IOException
+  {
+    long bytesRead = 0;
+
+    try
+    {
+      ensureProperFile(false, position);
+      bytesRead = sageFileChannel.transferTo(position - fileOffsets[currFileIndex], count, target);
+
+      if (bytesRead < count)
```
Narflex: You can just make this a 'while' instead of an 'if' and then remove the above 2 lines of code...then it will also cover spanning N number of files rather than just 2 (although, I don't think it'll ever be more than 2 in reality).

java/sage/media/bluray/BluRayFile.java
```diff
@@ -0,0 +1,496 @@
...
+  @Override
+  public FileLock tryLock(long position, long size, boolean shared) throws IOException
+  {
+    throw new IOException("Method is unsupported.");
+  }
+
+  @Override
+  public int read(ByteBuffer dst, long position) throws IOException
+  {
+    int readBytes = sageFileChannel.read(dst, position - fileOffsets[currFileIndex]);
```
Narflex: This looks like its missing a call to ensureProperFile(false, position) before the reads since the argument is an absolute position.

java/sage/media/bluray/BluRayFile.java
```diff
@@ -0,0 +1,496 @@
...
+      readBytes += sageFileChannel.read(dst);
+    }
+
+    return readBytes;
+  }
+
+  @Override
+  public int readUnsignedByte() throws IOException
+  {
+    return sageFileChannel.readUnsignedByte();
```
Narflex: This should have an ensureProperFile(false) at the beginning...and actually all read methods, even non-absolute ones need that in case they are called when the current file is at EOF.

Narflex:
Note to self...files reviewed:
build.gradle
MediaServer
Mpeg2Reader
Mpeg2Transcoder
NetworkClient
Version
SageFileSource
LocalSageFile
RemoteSageFile
EncryptedSageFile
BufferedSageFile
SageDataFile
SageInputStream
SageOutputStream
SageIOFileChannel
SageFileChannel
LocalFileChannel
RemoteFileChannel (with issues)
BufferedFileChannel (with minor issues)
BluRayFile (with minor issues)

OK, done reviewing all the new IO/NIO stuff, which was the hard part...I'll finish up the rest of it after I give you a chance to address my other comments.

java/sage/nio/RemoteFileChannel.java
enternoescape: Actually, I think you're wrong. I traced every step of this this through the debugger and it does exactly what it should be doing. When the first buffer has space and the next few buffers don't have anything, the offset is still 0. When the first buffer has nothing and the second buffer has space, the offset is returned as 1.

I  looked the problem so hard I went from not seeing it to actually creating a new problem in my last commit.

java/sage/nio/BufferedFileChannel.java
enternoescape: You're correct.

java/sage/media/bluray/BluRayFile.java
enternoescape: Changed it.

java/sage/media/bluray/BluRayFile.java
enternoescape: Agreed. I changed it to do that.

java/sage/media/bluray/BluRayFile.java
enternoescape: That's true. I think I  was just going with the general format of everything else.

java/sage/media/bluray/BluRayFile.java
enternoescape: Fixed it.

java/sage/media/bluray/BluRayFile.java
enternoescape: Fixed it.

java/sage/media/bluray/BluRayFile.java
enternoescape: Fixed it and adjusted the count according to what's remaining to be read so we don't loop endlessly.

java/sage/nio/RemoteFileChannel.java
enternoescape: In the latest commit, this is working correctly. Once it finds a buffer with space in it, it sets bufferAvailable to true and the offset ceases to increment even if the next buffer is empty.

java/sage/nio/BufferedFileChannel.java
enternoescape: Fixed it.

java/sage/nio/RemoteFileChannel.java
enternoescape: Changed things slightly, but according to several different scenarios I ran through the debugger, this code is working exactly as expected.

java/sage/media/bluray/BluRayFile.java
enternoescape: I fixed this to work similar to the other methods that provide a position to start reading from.

java/sage/media/bluray/BluRayFile.java
enternoescape: All of the NIO methods at worst will return 0 or -1. They should only be throwing an exception if there is really an exception and not for EOF. I guess it would be good to not waste a read that effectively returns nothing when it could have been prevented. I was just thinking it's more efficient to only check if we need to switch to a new file if the read returns 0 or -1 (depending on the documentation for the specific method), but maybe there's just as much code involved in either path.

enternoescape:
I can say the tests are doing their job. That's how I just caught a not very obvious bug in transferFrom.

I think I'm up to speed on your suggestions and addressed the few bugs. I have to say I really enjoy actually having someone critically review my code. :)

java/sage/io/RemoteSageFile.java
```diff
@@ -0,0 +1,599 @@
...
+
+    // End of the file.
+    if (len == 0)
+      return -1;
+
+    try
+    {
+      outStream.write(("READ " + remoteOffset + " " + len + "\r\n").getBytes(Sage.BYTE_CHARSET));
+      inStream.readFully(b, off, len);
+      outStream.flush();
```
Narflex: This flush() call and the one 7 lines down need to be moved up before you read from the stream.

java/sage/nio/RemoteFileChannel.java
Narflex: Yeah, sorry, apparently I missed how bufferAvailable was working. :)

java/sage/media/bluray/BluRayFile.java
```diff
@@ -0,0 +1,513 @@
...
+
+  @Override
+  public long skip(long n) throws IOException
+  {
+    if (n < 0)
+      return 0;
+
+    long seek = Math.min(position() + n, totalSize);
+    position(seek);
+    return seek - position();
```
Narflex: This looks like it will always return zero now.

Narflex:
This is about the most intensive code review I've done in awhile...overall, it's been looking really good. Let's see how much more I can get through now...maybe I'll be able to finish it today.

java/sage/FileDownloader.java
```diff
@@ -239,7 +241,7 @@ public synchronized Object downloadFile(String serverName, String srcFile, java.
           }
           break;
         }
-        fileOut = new FastRandomFile(myDestFile, "rw", Sage.I18N_CHARSET);
+        fileOut = new LocalSageFile(myDestFile, false);
```
Narflex: FastRandomFile does buffered writes, so you should probably wrap this in the buffering class for equivalence.

java/sage/FileDownloader.java
```diff
@@ -318,7 +320,7 @@ else if (remoteUIXfer)
         outStream.flush();
         str = Sage.readLineBytes(inStream);
         remoteSize = Long.parseLong(str.substring(0, str.indexOf(' ')));
-        fileOut = new FastRandomFile(myDestFile, "rw", Sage.I18N_CHARSET);
+        fileOut = new LocalSageFile(myDestFile, false);
```
Narflex: FastRandomFile does buffered writes, so you should probably wrap this in the buffering class for equivalence.

java/sage/IOUtils.java
```diff
@@ -1405,18 +1408,64 @@ public static String convertSMBURLToUNCPath(String smbPath)
...
+    java.io.InputStream fis = null;
     try
     {
-      fis = new java.io.FileInputStream(filePath);
+      if (local)
+        fis = new java.io.FileInputStream(filePath);
+      else
+      {
+        // This is crucial. If we don't do this step the file will almost certainly be inaccessible.
+        sage.NetworkClient.getSN().requestMediaServerAccess(filePath, true);
```
Narflex: Can you add a finally{} clause which released the media server access so the HashSet on the server doesn't grow without bound?

java/sage/media/bluray/BluRayFile.java
enternoescape: That's such a bad mistake. I can't believe I pushed that.

enternoescape: That's such a bad mistake, I can't believe I pushed that.

java/sage/media/sub/SubtitleHandler.java
```diff
@@ -121,7 +134,11 @@ else if (rv.length() == 0 && willDisplay)
         }
       }
     }
-    //System.out.println("Got Subtitle text=" + rv + " subEntryPos=" + subEntryPos);
+    finally
+    {
+      subtitleLock.readLock().unlock();
+    }
+    System.out.println("Got Subtitle text=" + rv + " subEntryPos=" + subEntryPos);
```
Narflex: You may want to get rid of this debug line...I'd imagine it logs a LOT when you do subtitle playback. You may also want to add some kind of "debug_subs" property which controls logging in the subtitle classes too.

java/sage/media/sub/SubtitleHandler.java
```diff
@@ -341,15 +372,22 @@ else if (sage.media.format.MediaFormat.VOBSUB.equals(targetFormat))
...
         subEntryPos = -1;
       }
-      return insertEntryForPostedInfo(time, dur, rawText);
+      finally
+      {
+        subtitleLock.writeLock().unlock();
+      }
     }
+
+    return insertEntryForPostedInfo(time, dur, rawText);
```
Narflex: This *should* be OK, but you may want to put the return statement before the finally block in order to match more closely the prior behavior.

Narflex: Oh wait...that won't work because then you'll be asking for the lock again...could work around that with a parameter, but it's likely fine the way you have it.

Narflex: Take back my take back since now I noticed you are using a re-entrant lock. :)

java/sage/media/sub/VobSubSubtitleHandler.java
```diff
@@ -46,24 +54,25 @@ public void loadSubtitlesFromFiles(sage.MediaFile sourceFile)
     try
     {
       subFile = getLoadableSubtitleFile(sourceFile, new java.io.File(subs[0].getPath()));
-      inStream = sage.IOUtils.openReaderDetectCharset(subFile, sage.Sage.BYTE_CHARSET);
+      inStream = sage.IOUtils.openReaderDetectCharset(subFile, sage.Sage.BYTE_CHARSET, sourceFile.isLocalFile());
```
Narflex: Looks like you forgot to update this one to not use getLoadableSubtitleFile (and also set the subFile path the way you were doing in the other classes).

java/sage/media/sub/VobSubSubtitleHandler.java
```diff
@@ -46,24 +54,25 @@ public void loadSubtitlesFromFiles(sage.MediaFile sourceFile)
...
       }
       String orgSubPath = subs[0].getPath();
-      bitmapFile = getLoadableSubtitleFile(sourceFile, new java.io.File(orgSubPath.substring(0, orgSubPath.length() - 3) + "sub"));
-      if (bitmapFile == null)
+      bitmapFile = new java.io.File(orgSubPath.substring(0, orgSubPath.length() - 3) + "sub");
+      if (!sourceFile.isLocalFile())
       {
-        if (sage.Sage.DBG) System.out.println("Invalid VobSub; missing the .sub file for .idx file: " + orgSubPath);
-        return;
+        sage.NetworkClient.getSN().requestMediaServerAccess(bitmapFile, true);
```
Narflex: You should release this request for access when this is closed if possible.

java/tv/sage/ModuleGroup.java
```diff
@@ -293,27 +293,32 @@ else if (cryptoKeyObj instanceof Object[])
         }
       }
       file.createNewFile();
-      sage.FastRandomFile widgetDBout = new sage.FasterRandomFile(file, cryptoKey == null ? "rwd" : "rwdc", sage.Sage.I18N_CHARSET);
-      if (cryptoKey != null) widgetDBout.setCrypto(cryptoKey);
+      sage.io.SageDataFile widgetDBout;
+
+      if (cryptoKey == null)
```
Narflex: Feel free to get rid of the cryptoKey stuff in this class...that was only ever used with SageTVLite when we encrypted the STV for it to block access to extra functionality.

Narflex: Lol...you can still remove it...but apparently it wasn't just SageTVLite, I think we actually encrypted all the .stv files originally until we switched to XML files. Geeze...I was paranoid. :)

native/exe/SageLauncher/SageLauncher.rc
```diff
@@ -69,12 +69,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "SageTV Open Source"
             VALUE "FileDescription", "SageTV"
-            VALUE "FileVersion", "9.0.4.232"
+            VALUE "FileVersion", "9.0.5.232"
```
Narflex: I think this one should be 9.0.5.287

native/exe/SageLauncher/SageLauncher.rc
```diff
@@ -69,12 +69,12 @@ BEGIN
...
             VALUE "InternalName", "SageTV"
-            VALUE "LegalCopyright", "Copyright � � 2015 The SageTV Authors"
+            VALUE "LegalCopyright", "Copyright � � 2015 The SageTV Authors"
```
Narflex: While you're in here...why don't you make it 2015-2016.

java/sage/media/sub/SRTSubtitleHandler.java
```diff
@@ -159,4 +444,67 @@ else if (!inTag)
...
 
+  @Override
+  public long getTimeTillUpdate(long currMediaTime)
+  {
+    long next = super.getTimeTillUpdate(currMediaTime);
+
+    if (monitorThread == null)
+      return next;
+
+    if (next == sage.Sage.MILLIS_PER_WEEK)
```
Narflex: I'd suggest changing this to be another constant such as NO_MORE_SUBS_LONG_WAIT and changing the other places its used in the subtitle classes to match that, also makes the interdependence more clear.

Narflex:
OK, I've now read through everything and you can see the remaining comments I had. Great work overall!  I didn't go through the updated SRTSubtitleHandler with a fine-toothed comb, but I'm sure you've been using it extensively so it's likely working fine.

And I remembered another good reason to increment the version...then it'll back up the Wiz.bin file when people update to this version...and since you've changed the file handling code for that, we definitely want people to have a backup of it. :)

java/sage/io/RemoteSageFile.java
enternoescape: Fixed it and I noticed that I did something less than efficient with readFully, so I fixed that too.

java/sage/IOUtils.java
enternoescape: That makes sense.

java/sage/FileDownloader.java
enternoescape: Done.

java/sage/FileDownloader.java
enternoescape: Done.

java/sage/media/sub/SubtitleHandler.java
enternoescape: Good idea. I actually forgot that debug line was originally commented out. It was very helpful during debugging.

enternoescape: I went through all of the subtitle classes and applied the the subtitle debug boolean only to things that were not actual errors or warnings.

java/sage/media/sub/SubtitleHandler.java
enternoescape: I was trying to remove the need to lock at that spot if it wasn't needed. I took at a look at how the parameters are actually used and determined that we didn't need to get the lock, but to make things closer to as they were I changed it per your suggestion.

java/sage/media/sub/VobSubSubtitleHandler.java
enternoescape: I spotted that too. It's crazy. You'd think I didn't look this over a few times before I pushed it.

java/sage/media/sub/VobSubSubtitleHandler.java
enternoescape: Once the file is opened, remove access only keeps it from being re-opened correct?

enternoescape: I think I just answered my own question. It looks like it only keeps you from re-opening the file. Which is ok as long as you continue to access the file under 1.5 minutes. I'll need to think about this. Maybe it actually does make more sense to have the remote file classes get and release access automatically.

java/sage/media/sub/SRTSubtitleHandler.java
enternoescape: That's a good idea. Done.

native/exe/SageLauncher/SageLauncher.rc
enternoescape: Done.

native/exe/SageLauncher/SageLauncher.rc
enternoescape: Done.

java/sage/media/sub/VobSubSubtitleHandler.java
enternoescape: I found a way around the problem.

java/tv/sage/ModuleGroup.java
enternoescape: I saw that the only thing using it was the method that basically passes null for the encryption key. Per your suggestion, it's now gone. I did occasionally run into your comments about paranoia in the code; at times I get a good laugh. :)

enternoescape: This confused me a little:
`widgetDBout.writeUnencryptedByte(cryptoKey != null ? (byte) 0x36 : (byte)0x20); // < 0x2F to signify unencrypted`

The code clearly shows 0x20, but your comment shows 0x2F. I presume the code is correct since the comment is just a comment. Should I update the comment?

enternoescape:
Backups are always nice.

Just an FYI. I will probably not be active for tomorrow and the weekend, so if you're waiting for input whereby silence means acceptance, please wait for Monday before you commit. I'll be essentially in the wilderness which is probably a good thing for me. I spend far to much time coding and browsing the internet.

Narflex:
Perfect. I have family in for the weekend so I wasn't going to get around
to anything until probably Tuesday. :-)

Jeff Kardatzke

enternoescape:
My thing is very similar, but it's more like a very large cabin with all of my extended family in the woods.

java/sage/media/sub/VobSubSubtitleHandler.java
```diff
@@ -40,30 +47,31 @@ public void loadSubtitlesFromFiles(sage.MediaFile sourceFile)
...
     java.io.BufferedReader inStream = null;
     getLanguages();
     try
     {
-      subFile = getLoadableSubtitleFile(sourceFile, new java.io.File(subs[0].getPath()));
-      inStream = sage.IOUtils.openReaderDetectCharset(subFile, sage.Sage.BYTE_CHARSET);
+      inStream = sage.IOUtils.openReaderDetectCharset(subFile, sage.Sage.BYTE_CHARSET, sourceFile.isLocalFile());
       String line = inStream.readLine();
-      if (line != null && line.indexOf("VobSub index file") == -1)
+      if (line != null && line.contains("VobSub index file"))
```
Narflex: The second part of the conditional is inverted over what it originally was.

java/tv/sage/ModuleGroup.java
Narflex: Leave the code the way it is, since that's likely correct. Feel free to update the comment. :)

Narflex:
OK, had one error I found and replied to another comment. It's almost ready for merging. :)

Narflex:
One other thing I noticed is that the test code is in /src/test/java, which is somewhat confusing because it has that 'src' root directory...but the other source code is not in that directory. What do you think about changing that to just be 'test/java' instead?

java/sage/media/sub/VobSubSubtitleHandler.java
enternoescape: Not sure why I did that. Fixed it.

enternoescape:
That's fine. I originally did test/java, but it felt kind of wrong. I'll change it back to that.

Narflex:
Woo-hoo! Congrats! It's merged!

PR #160, opened 2016-07-25
Fix omitted bufferSize from START command parsing

CraziFuzzy:
Regarding Issue #159


Narflex:
Yup, that's correct. I can see that change was removed from between the V7 code and the open source version...I have no idea why though. :) (I looked, and I think it was from somebody else doing a code cleanup who likely removed the line due to it being an unused variable)

PR #175, opened 2016-09-17
Refactored GSON 2.7 for Schedules Direct

enternoescape:


Narflex:
I just need to check with some folks internally to see if this needs to go in the third_party directory or not since it is Google code, but from a different project.

enternoescape:
I can just move it and save you the effort. It might make it easier to understand what's going on down the road too.

Narflex:
I already asked....your call, I'll get a response but it might not be until
Monday.

Jeff Kardatzke

enternoescape:
I moved it. Let me know if you need further details in the folder or what other files need to be updated for legal reasons.

CraziFuzzy:
So will we also need to alter the plugin repo to remove the gson.jar with this change? (I'm reading this as pulling GSON into the core, right?)

enternoescape:
It's refactored into another package, so it will not conflict with GSON if someone else imports a different version of GSON for a plugin.

CraziFuzzy:
Got ya.. if I had thought for more than a couple seconds on it I would have realized that.. ;-)

PR #176, opened 2016-09-18
Schedules Direct POJO's, connectivity and some utility classes.

enternoescape:
In IOUtils.java, I specified 32 characters for the MD5 hash StringBuilder method since that's the size it will always be and I was making a similar small optimization on the SHA1 hash method. My reasoning is the char array default is 16 will get re-created at least once and why would we allow that when we know what the final size will always be.

The POJO's as you might expect very closely resemble the JSON returned from Schedules Direct and in cases where it made sense, I added helper methods or intermediate processing that generally is always desirable. Most cases where an array is returned, there is an assurance that at the very least an empty array will be returned (not null). They are all under sage.epg.sd.json and in some ways exceed the amount of information we actually are using, but even the things we don't use get decoded either way, so why not make them available just in case we want to use them later.

SDImages.java is used to "encode" and "decode" the images into a single byte array for channel logos and a 2D array for Shows and SeriesInfo.

SDUtils.java is used to convert between what's native for Schedules Direct and SageTV, plus some "filters" to clean things up a little.

SDSession.java is an abstract class that just needs a connectivity class (SDSageSession.java) to perform all of the HTTPS related interactions with Schedules Direct. It also manages most of the errors we might receive and will automatically retry once for token related issues or other situations where an immediate retry might make sense. Otherwise it will try to throw the most appropriate exception. It also contains optional logging of all interactions between SageTV and SD to the level of what exactly was sent and what exactly was returned.

There are also several related tests, but I will hold off on those until the next push.


Narflex:
I only really reviewed the code that was changing the existing stuff in the SageTV core. The rest I just glossed over because it clearly can't break anything in the current system because it's not being referenced yet. :)

enternoescape:
Haha. I hope this will make it easier to see the things you actually would be interested in which will be in the next commit.

PR #177, opened 2016-09-20
Schedules Direct EPG grabber and associated STV and API changes.

enternoescape:
The main part of this commit is SDRipper.java which handles getting a processing EPG data from Schedules Direct. There are also a few more API changes to more generically support EPG. The STV has also been updated to reflect the newly available EPG service.

The other less related changes are to the old EPG classes mostly to make them easier to read and less mistake-prone when it comes to generic types.


enternoescape:
There's definitely more that can be done with this EPG source, but this is probably enough for most people to get what they are looking for today.

enternoescape:
Something isn't working regarding dates. I'll push again after I figure out what is happening.

enternoescape:
I have it working now. There's no sense in creating a new request.

java/sage/Channel.java
```diff
@@ -74,6 +76,14 @@
     }
     else
       logoImages = Pooler.EMPTY_INT_ARRAY;
+    if (ver >= 0x56)
+    {
+      int urlLen = in.readShort();
+      logoURL = new byte[urlLen];
```
Narflex: In the one arg constructor, set logoURL to be an empty byte array. This shouldn't matter, but it's a good idea for consistency.

java/sage/EPG.java
```diff
@@ -1324,38 +1329,38 @@ public int getServiceLevel(long providerID, int stationID)
...
       int idx = 0;
       while (walker.hasNext())
-        rv[idx++] = ((Integer) walker.next()).intValue();
+        rv[idx++] = walker.next().intValue();
       return rv;
     }
   }
 
-  private static java.util.Map parseChanMapString(String str)
+  private static java.util.Map<Long, Integer> parseChanMapString(String str)
```
Narflex: You changed this from being an Integer, Integer map to being a Long, Integer map. It should be an Integer, Integer map because it's mapping station IDs to service levels and station IDs are Integer values....there's a few spots where this needs to be changed to resolve that.

java/sage/EPG.java
```diff
@@ -1631,25 +1662,25 @@ public long getCachedProviderIDForName(String s)
...
 
   private Object stationIDCacheLock = new Object();
   private int[] viewableStationIDCache;
 
-  private java.util.Map lineups;
-  private java.util.Map lineupOverrides;
-  private java.util.Map serviceLevels;
+  private final java.util.Map<Long, java.util.Map<Integer, String[]>> lineups;
+  private final java.util.Map<Long, java.util.Map<Integer, String[]>> lineupOverrides;
+  private final java.util.Map<Long, java.util.Map<Long, Integer>> serviceLevels;
```
Narflex: You changed this from being an Integer, Integer map to being a Long, Integer map. It should be an Integer, Integer map because it's mapping station IDs to service levels and station IDs are Integer values....there's a few spots where this needs to be changed to resolve that.

enternoescape: I made it Integer, but then you had a line that was importing it at long and then I thought that's what it must be.

java/sage/EPGDataSource.java
```diff
@@ -636,8 +638,8 @@ protected boolean extractGuide(long guideTime)
     {
       // We still need to put empty maps in there so it thinks there's actually lineup
       // data for this source...and there is in the overrides.
-      EPG.getInstance().setLineup(providerID, new java.util.HashMap());
-      EPG.getInstance().setServiceLevels(providerID, new java.util.HashMap());
+      EPG.getInstance().setLineup(providerID, new java.util.HashMap<Integer, String[]>());
+      EPG.getInstance().setServiceLevels(providerID, new java.util.HashMap<Long, Integer>());
```
Narflex: Like the EPG.java change, this should be an Integer, Integer map.

java/sage/EPGDataSource.java
```diff
@@ -656,7 +658,7 @@ protected boolean extractGuide(long guideTime)
       lineMap.put(new Integer(defaultStationID + i), new String[] { Integer.toString(i) });
     }
     EPG.getInstance().setLineup(providerID, lineMap);
-    EPG.getInstance().setServiceLevels(providerID, new java.util.HashMap());
+    EPG.getInstance().setServiceLevels(providerID, new java.util.HashMap<Long, Integer>());
```
Narflex: Again, Integer-Integer map.

java/sage/WarlockRipper.java
```diff
@@ -457,9 +491,9 @@ protected boolean extractGuide(long inGuideTime)
       }
 
       java.util.Set logosToDownload = new java.util.HashSet();
-      java.util.Map lineMap = new java.util.HashMap();
-      java.util.Map serviceMap = new java.util.HashMap();
-      java.util.Map physicalMap = null;
+      java.util.Map<Integer, String[]> lineMap = new java.util.HashMap<Integer, String[]>();
+      java.util.Map<Long, Integer> serviceMap = new java.util.HashMap<Long, Integer>();
```
Narflex: Again, change to Integer-Integer map

java/sage/WarlockRipper.java
```diff
@@ -510,7 +544,7 @@ protected boolean extractGuide(long inGuideTime)
         }
 
         lineMap.put(new Integer(stationID), chanNums);
-        serviceMap.put(new Integer(stationID), new Integer(serviceLevel));
+        serviceMap.put(new Long(stationID), new Integer(serviceLevel));
```
Narflex: Again, change to Integer-Integer map

java/sage/Wizard.java
```diff
@@ -3156,6 +3177,11 @@ public SeriesInfo addSeriesInfo(int legacySeriesID, int showcardID, String title
         series.seriesImages = Pooler.EMPTY_INT_ARRAY;
       else
         series.seriesImages = seriesImages.clone();
+      if (imageURLs == null || imageURLs.length == 0)
+        series.imageURLs = Pooler.EMPTY_2D_BYTE_ARRAY;
+      else
+        series.imageURLs = imageURLs;
+      series.imageURLs = (imageURLs == null) ? Pooler.EMPTY_2D_BYTE_ARRAY : imageURLs.clone();
```
Narflex: You're setting series.imageURLs twice here. Just remove this last line and then add a .clone() on the above line (yeah, I know it's not fully consistent about cloning array arguments on all the object creations, I used to be paranoid but got less paranoid as time went on)

java/sage/EPG.java
enternoescape: Either way, I fixed them.

java/sage/Wizard.java
enternoescape: Good point. Fixed it.

I tried my best to emulate whatever you were doing for the particular object I was working on.

Narflex:
Note to self....I've reviewed all of the changes to the SageTV core Java files and made a few comments (mostly which seem addressed already, groovy).  I'm not going to review the SDRipper in detail because that'd take too much time getting an understanding of how their data feed works and such. I'll leave that to beta testing by users. :) (but IMO we won't need a beta version since it's an isolated component, it's just a beta feature)

Tomorrow I'll try to review the STV changes, and if those are fine then I'll merge it. Nice work!

enternoescape:
Yay!

Narflex:
I took a look at the STV and there's a few changes that don't seem like they were needed and were likely just accidents when you were moving stuff around. You should do a UID diff against the starting SageTV7.xml, find these in the list, right click and choose 'Revert Differences' from that menu and it should then properly undo them.  Unless you meant to do any of these...and if so, let me know why. :)

1. Removed Left-Back from Theme SetupWizardMenuTheme 7
2. Added Setup Wizard - Schedules Direct Country Selection to the above theme
3. Removed Theme SetupWizardMenuTheme 7 from Menu Setup Wizard - Lineup: check to reuse or copy
4. Removed Listener Left-Back from Menu Setup Wizard - Lineup: check to reuse or copy
5. Removed Attribute MenuTitle from Menu Setup Wizard - Lineup: check to reuse or copy

If you go through all the diffs one by one, these are the only ones where you actually changed the original STV rather than just adding new isolated menus (with the exception of rewording the text on where you select which EPG service you are using, and adding in the button for SD there and the code that hangs off that).

enternoescape:
Will do. I also just added the small changes you indirectly suggested in the issue thread.

enternoescape:
There is no indication in the API that I could find that shows that's a possibility, but I checked my JSON logs from my last sync and I see it being used there. I'll make the changes.

enternoescape:
Thanks for the pointers. These small changes appear to have a big impact.

Narflex:
OK...that totally broke the STV. I can't even load it in the Studio anymore...I'm going to look into why that happened and likely push a fix for it, and then you can go back to the STV from before and then properly revert the changes.

Narflex:
I committed the fix to that problem. So just do what you did again after grabbing that fix (but of course do that operation on the STV before it got screwed up by the bug in the code)

enternoescape:
I double-checked this one before I uploaded it. I was able to re-open it in Studio.

Narflex:
Just a reinforcement...I did NOT review all of the SD specific code, just made sure you didn't break anything else in SageTV. Be sure to warn people it's beta code when they try it out.  And I also want the version updated as part of this...after I merge your change I'll just do that myself. :)

Narflex:
OK, version increment pushed. :)

PR #179, opened 2016-09-23
Added Schedules Direct to change log as a new beta feature.

enternoescape:
Almost forgot to update the change log.


enternoescape:
When I put the changes on my production machine I immediately noticed that I had not considered the possibility that the very last schedule added might not be unique and if there were less than 1000 unique schedules, then nothing would be updated.

stuckless:
FYI... If everyone is happy with the SD merge... I'll kick off a build this weekend (Likely saturday) and push out the 9.0.8 release with SD support.  Awesome work.

enternoescape:
After this gets merged, I'm OK with a new release. The issue I just addressed would definitely be noticed by many users and did not come up in testing I believe because I wasn't being selective about what channels I wanted to be able to see in the guide.

enternoescape:
This change is related to the first one and causes the same kind of problem. I haven't had it happen yet, but it could happen for similar reasons.

PR #180, opened 2016-09-23
Moved JSON strings in tests to .json files to improve compile times.

enternoescape:
It was really bothering me how I pieced the JSON deserializer tests together and it seemed to be taxing on the compiler. Moving these strings into .json files cut the start to finish times for building Sage.jar from 2 minutes to 1 minute.


enternoescape:
I was going to hold off on this, but it's a lot of files being changed that you don't care about and I didn't want to make a future push harder to follow.

Narflex:
I don't ever review the test code...so if it passes the build test, it's good enough for me. :)

PR #185, opened 2016-09-27
Fixed bug in the EPG license detection logic. We now also check for NOKEY.

enternoescape:
This is a real problem for people using XMLTV plugins because they keep getting error messages and the guide data is not being imported. Not surprisingly, this does not prevent the plugin itself from loading, but since we check for the license key before we try to use the plugin, it fails immediately.


Narflex:
Thanks!

PR #186, opened 2016-09-27
Fixed severe bug in Plugin Manager

stuckless:
that prevented files from being upgraded correctly.


Narflex:
Oh wow...don't know how we both missed that one before. :)

stuckless:
Yeah... that some pretty bad side effects :(  I did some stand alone testing as well upgrading a system over and over... and things appear to be working well, now.

PR #194, opened 2016-10-12
Enable streams with valid PAT packets, but invalid PMT packets to be detected.

enternoescape:


jason-bean:
I notice there is a enum value "PRIVAITE_DATA" in STREAM_DATA_TYPE that appears to be misspelled throughout the source. It only seems to affect 7 files. I might suggest correcting this misspelling before this pull request is merged.

Narflex:
There's lots is misspellings in the native code...feel free to fix this one
if you want, but I'm fine either way.

Jeff Kardatzke

jason-bean:
Ok, I'll do that after this is merged. I don't want to interfere with this pull.

PR #196, opened 2016-10-12
Fixed the line endings to accommodate different git configurations independent of the platform.

enternoescape:


test/java/sage/IOUtilsTest.java
```diff
@@ -14,14 +14,26 @@
   public void testGetUrlAsString() throws Exception
   {
     String md5 = IOUtils.getUrlAsString(file2url(getTestResource("test-jar-1.0.zip.md5")));
-    assertEquals(md5, "beec1608cf997d3acb42f0ab772b143f" + System.lineSeparator());
+    assertNotNull(md5);
```
Narflex: Why not just use String.trim() on the MD5 values and then compare them after doing that instead?

enternoescape: I guess that would have the same effect. I was thinking what if more than what we were expecting was read for some reason. Trim might remove a real problem.

Narflex:
Oh sorry...my bad. I thought it was doing an MD5 comparison because it was referencing MD5 values...I didn't realize this was testing getFileAsString and getURLAsString...in that case, then I'd think there should be no difference at all per-platform relating to line endings. I'll take a look at this and likely submit another change. :)

Narflex:
OK, the problem is that the GIT client may end up modifying the line endings on the test data file when it is downloaded...that's why it can have \r\n or \n as the line ending.  So now I actually prefer your original fix then. :)  I'll revert it back to what you originally submitted...my bad.

Narflex:
Reversion back to your original fix submitted.

enternoescape:
Now worries. Thanks for the laugh. :)

PR #200, opened 2016-10-24
Check for root to enable ThreadPriorityPolicy

cbasolutions:
Added $(if [[ $EUID -eq 0 ]]; then echo '-XX:ThreadPriorityPolicy=1'; fi) so that when SageTV is started by non-root users it will not throw an error due to lack of permissions.

### DIFF
--- a/build/serverfiles/startsagecore
+++ b/build/serverfiles/startsagecore
@@ -42,7 +42,7 @@ while [ $restartApp -eq 1 ]; do
 	fi
 	done
 	echo Starting server mode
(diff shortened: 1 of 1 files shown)


cbasolutions:
CLA was signed as CBA Solutions, LLC. GitHub updated to use same email.

Narflex:
It's not showing you as a signer when I use the Google internal tool to check. I'm also able to look them up by email...you can easily guess my Gmail address if you want to send me your email that way.

PR #213, opened 2016-11-14
Fixes a NPE that happens in some coditions

stuckless:
When the IP address can't be determined (ie, not using eth0 or eth1) you get a NPE.   This fixes the NPE, but also adds ent1 br0 and docker0 to the list of devices to check for IP address.  The code will iterate all those devices looking for the first one that returns an address.

### DIFF
--- a/java/sage/LinuxUtils.java
+++ b/java/sage/LinuxUtils.java
@@ -561,7 +561,7 @@ public class LinuxUtils
     "inet addr\\:(\\p{Digit}\\p{Digit}?\\p{Digit}?\\.\\p{Digit}\\p{Digit}?\\p{Digit}?\\.\\p{Digit}\\p{Digit}?\\p{Digit}?\\.\\p{Digit}\\p{Digit}?\\p{Digit}?) ");
 
   static String getIPAddressFromInetInfo(String inetInfo) {
-    if (inetInfo.contains("UP"))
+    if (inetInfo!=null && inetInfo.contains("UP"))
     {
       java.util.regex.Matcher mat = INETINFO_IP_PATTERN.matcher(inetInfo);
       // Go with eth1
--- a/java/sage/LinuxUtils.java
+++ b/java/sage/LinuxUtils.java
@@ -15,6 +15,10 @@
  */
 package sage;
 
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


enternoescape:
This is actually even more complicated than this. A lot of modern Linux OS's have biosdevname enabled by default which causes the interfaces to be named either based on things like what slot they are plugged into or the name the manufacture wanted, etc.

Outside of Linux, we are ok with using the Java API to get the IP addresses of the adapters, why is that not sufficient for Linux too? I see the comments about why we get the address this way, but if the Windows and Mac approaches work without any real problems, why is Linux special?

stuckless:
I'm not 100% why it couldn't be a pure java call, except, and I haven't validated doing this in Java.  My intention here was to simply to restructure this to avoid the NPE and to ensure that it would work better, accounting for more linux device names.  (On my docker setup, it would always getting a NPE because eth0|1, which it exists, doesn't have an IP).  The code also allows you to set in the Sage.properties what your wired eth device would be called, so, if all else fails, at least have you a way to tell sagetv which device to obtain it's IP Address.

I do know that I've read lots about the Java apis for getting an IP address is very unreliable... but, as of Java 7 or later, maybe that has changed.

enternoescape:
Understood. I was more directing that question @Narflex to see what the original reason was for going this route. I didn't need to go this route with any of the other software I wrote, so I was just wondering if things were really that bad in Java at the time this was done.

It's also worth noting that `ifconfig` is a deprecated command and `ip` is taking over. Many ditros either do not install net-tools by default or dropped that package completely in favor of iproute2. This is more of a future note that we might need to revisit this.

stuckless:
I think it's worth revisiting the Java Network APIs and seeing if we could do this entirely in java... I don't really have the time, right now, to do it, but it should be done.  The more "pure" java calls we make the more portable the the code becomes.

Narflex:
The original reason we did this was because on some platform many years ago we were supporting (I don't recall which one, likely some OEM deal)...the Java API was returning 0.0.0.0 for this (or maybe 127.0.0.1)...so we worked around it by using this ugly method.  I'd really likely to change this...it sucks doing it this way...I'm also unsure about the current state of the Java API and what it will return for this case.  Maybe somebody can clean it up more in the future.

PR #221, opened 2016-11-23
Updated DVB-S & DVB-T frequencies for New Zealand

JREkiwi:
Updates for restack of NZ DVB-T in 2014
Updates to add all Optus DVB-S transponders avilable in New Zealand
Create separate definition for Optus D1 for New Zealand only beam to cater for future addition of Australia only beam


JREkiwi:
|I signed it!|

Issue #146, opened 2016-06-24
Server does not stay awake for Transcode Jobs


CraziFuzzy:
Need to add logic to ServerPowerManagement.java to check if Ministry has any jobs, and if it does, flat the system to stay on.

CraziFuzzy:
Quick and simple changes made and pull request made. #148

CraziFuzzy:
This may need a tweak.  I overlooked that completed and failed jobs remain in the list, so using **GetTranscodeJobIDs()** is still holding jobs even if nothing is converting or waiting to convert.  I need to change **Ministry.requiresPower()** to check **waitingForConversion.size()** and **converting.size()** instead (thus ignoring jobs held in **waitingForAbsolution** )

PR #223, opened 2016-11-30
Fixed so that Ministry will not allow sleep while converting.

CraziFuzzy:
As per #146, should now only look for queued and in progress jobs to prevent sleep.  Previously, any job in the list, including failed and completed jobs, were preventing sleep.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Version 9.0.10 (??)
 * Fix: Enable streams with valid PAT packets and invalid PMT packets to be able to be detected by the built in remuxer.
 * New: Schedules Direct now includes teams as people for favorite scheduling.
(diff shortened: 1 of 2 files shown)


CraziFuzzy:
Well, aren't YOU speedy today @Narflex. :-)

TuscanRaider:
Does this code still need a synchronize block?

CraziFuzzy:
Probably not exactly needed - but I didn't go through the entire code looking for potential times where the collections it's looking at might be mid-alteration.  It looks like things are just removed from one collection and added to another, so it'd probably be fine to remove it.  That said, this is a ridiculously short and infrequent call that wouldn't really affect Ministry in it's tasks, and wouldn't really be affected being blocked  if Ministry is busy when it's been called.

PR #224, opened 2016-12-01
Fix to allow mounting DVD iso images as non-root

stuckless:
This allows a non-root instance of sagetv to import a dvd iso (previously it would fail).   I'm unsure if there are other changes that would be required in order to actually play the dvd... but these changes do allow it to be imported.

There are many places in SageTV where "mount" is used... I'm thinking that a better long term solution might be to create a "Mount" class that manages the need for sudo, and then we find/replace all the exec "mount" (and "umount") places with a Mount class than can manage this.

### DIFF
--- a/java/sage/FSManager.java
+++ b/java/sage/FSManager.java
@@ -416,7 +416,7 @@ public class FSManager implements Runnable
           // mount requires root permissions
           // the sagetv user needs to be in the sudoers list
           // NOTE mount command is NOT resolved in properties because sudo is being used
-          if (IOUtils.exec2(new String[]{"sudo","mount", isoFile.getAbsolutePath(), mountDir.getAbsolutePath(), "-o", "loop,ro"}, true) != 0)
+          if (IOUtils.exec2(new String[]{"sudo","-n", "mount", isoFile.getAbsolutePath(), mountDir.getAbsolutePath(), "-o", "loop,ro"}, true) != 0)
           {
             if (Sage.DBG) System.out.println("FAILED mounting ISO image " + isoFile + " to " + mountDir);
             return null;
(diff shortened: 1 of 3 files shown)


java/sage/FSManager.java
```diff
@@ -478,7 +492,12 @@ public void releaseISOMount(java.io.File mountDir)
         }
       }
       else
-        IOUtils.exec2(new String[] { "umount", mountDir.getAbsolutePath() }, false);
+      {
+        if (Sage.LINUX_IS_ROOT)
+          IOUtils.exec2(new String[]{"umount", mountDir.getAbsolutePath()}, false);
+        else
+          IOUtils.exec2(new String[]{"sudo","umount", mountDir.getAbsolutePath()}, false);
```
enternoescape: This will hang indefinitely waiting for a password if the sudoers configuration is not set to not require a password for at least this command.

enternoescape: To clarify, it will timeout eventually, but it could take about 5 minutes or more depending on the system defaults. In the mean time a few threads are tied up for each mount attempt.

enternoescape: Sorry, I saw the commit code before I saw the comments. I guess since root is the default configuration a user would actively be trying to run SageTV as a non-root user to see this kind of problem crop up. My emphasis was not so much that the user needs to be on the list so much as the command also need to be set to not require a password too, but both are important.

stuckless: I updated sudo to use -n (non interactive) so that it will fail fast if a password is required.

```
Thu 12/1 12:26:08.888 [LibraryImporter@6b59f53b] Executing process: [sudo, -n, mount, /var/media/videos/American Sniper.iso, /tmp/formatmount, -o, loop,ro]
Thu 12/1 12:26:08.888 [LibraryImporter@6b59f53b] Started process object: java.lang.UNIXProcess@659e1620
Thu 12/1 12:26:08.890 [ErrorStreamConsumer@1408e2a5] STDERR:java.lang.UNIXProcess@659e1620: sudo: a password is required
Thu 12/1 12:26:08.891 [LibraryImporter@6b59f53b] FAILED mounting ISO image /var/media/videos/American Sniper.iso to /tmp/formatmount
```

PR #237, opened 2017-01-04
Added better handling to Seeker when starting a recording and no directories are selectable for the desired encoder.

enternoescape:
Instead of just returning what would effectively be a message saying there was a null pointer exception (not helpful at all to normal users really), when we know it will be thrown because bestStore is null, we clean up a little  and throw a different exception earlier. The message in the thrown exception still doesn't point you to exactly where the problem is, but it should make you think about what you last changed.

I also fixed a small typo I happened to notice.


enternoescape:
This resolves #236.

PR #247, opened 2017-01-11
Enable all local UI's to configure placeshifter

CraziFuzzy:
regarding issue #246

### DIFF
--- a/java/sage/api/Global.java
+++ b/java/sage/api/Global.java
@@ -1716,6 +1716,22 @@ public class Global {
         }
         return Boolean.FALSE; // this means we're a full SageTV UI running under Linux
       }});
+    rft.put(new PredefinedJEPFunction("Global", "IsLocalConnection")
+    {
+      /**
+       * Returns true if this UI is on the server's subnet.
+       * @return true if this UI is on the server's subnet, false otherwise.
+       *
+       * @since 9.0.13
+       * @declaration public boolean IsLocalConnection();
+       */
+      public Object runSafely(Catbert.FastStack stack) throws Exception{
+        UIManager uiMgr = stack.getUIMgr();
+        if (uiMgr.getUIClientType() == UIClient.REMOTE_UI)
+          if (uiMgr.getRootPanel().getRenderEngine() instanceof MiniClientSageRenderer)
+            return ((MiniClientSageRenderer) uiMgr.getRootPanel().getRenderEngine()).isLocalConnection();
+        return Boolean.TRUE;
+      }});
     rft.put(new PredefinedJEPFunction("Global", "IsTouchUI")
     {
       /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Force debug logging to always be on.
 * Fix: Watched calculation for movies with commercials is improved
 * Fix: Prevent freezing between programs when playing back on Windows (matches V7 behavior, although not ideal, avoids freezing)
+* New: Add new API function Global.isLocalConnection() to identify UI's running within the server's subnet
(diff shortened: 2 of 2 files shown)


java/sage/api/Global.java
```diff
@@ -1716,6 +1716,22 @@ public Object runSafely(Catbert.FastStack stack) throws Exception{
...
+      /**
+       * Returns true if this UI is on the server's subnet.
+       * @return true if this UI is on the server's subnet, false otherwise.
+       *
+       * @since 9.0.13
+       * @declaration public boolean IsLocalConnection();
+       */
+      public Object runSafely(Catbert.FastStack stack) throws Exception{
+        UIManager uiMgr = stack.getUIMgr();
+        if (uiMgr.getUIClientType() == UIClient.REMOTE_UI)
```
Narflex: You can collapse these into a single conditional statement.

stvs/SageTV7/SageTV7.xml
```diff
@@ -67182,7 +67182,7 @@
           </Action>
           <Action Name="&quot;REM Option for: Configure Placeshifter Connection&quot;" Sym="OPUS4A-210962">
            <Conditional Name="!IsEmbeddedSystem()" Sym="OPUS4A-211270">
-            <Conditional Name="( IsServerUI() || (!IsDesktopUI() &amp;&amp; IsRemoteUI()) || (IsLinuxOS() &amp;&amp; GetServerProperty(&quot;linux/enable_nas&quot;,false)) ) &amp;&amp; IsSageTVServerEnabled()" Sym="BASE-45755">
+            <Conditional Name="IsLocalConnection() &amp;&amp; IsSageTVServerEnabled()" Sym="BASE-45755">
```
Narflex: The problem with this is that it will also return true for SageTVClient. And the UPnP automatic configuration won't work if run through SageTVClient on the same LAN. The reason is that in the STV code, it actually ends up getting the local IP address and then using that to forward the ports...it does this in the UI code so it can give feedback to the user about specific failures setting up UPnP.  So if you used SageTVClient, it would forward the traffic to the client and not the server...this would get fixed after a little while when the server notices the new property settings and then fixes the mapping...but would fail when the user tries to test the locator connection before that occurs. If you put it back the way it was and just change the !IsDesktupUI() part to be IsLocalConnection() then you won't have that problem.  Also, by having IsServerUI() in there, it will allow SageTVClient running on the same machine as SageTV server to configure it  (and I just noticed it would also allow placeshifter running on the same machine as the server to configure it as well).

CraziFuzzy: Okay, I can fix that I think by checking:
`( IsServerUI() || ( IsRemoteUI() && IsLocalConnection() ) ) && IsSageTVServerEnabled()`

I noticed the CLA check isn't working - it appears github isn't associating the commits from my newer dev machine with my CLA's login.  And tips on that?

Narflex: That change looks good then. And the CLA thing I believe is because it's using a different GitHub account. The one's here lists you as CraziFuzzy...whereas on the individual commits it shows you as 'christopher piper'. I know they're both you, so I'll merge it...but it may block me when I try to do that.

CraziFuzzy: well, since I've got to make changes anyway, I can just revert out the commits, and resubmit them under the 'right' account (strange, because I was pretty sure git was logging in as CraziFuzzy).  Either way, I've got to get it sorted in my git config, so I'll just do that.

jwittkoski: @CraziFuzzy It's probably looking at the author field in the commit and not who you were logged into when you pushed the change to github.

Narflex:
This is stale so I'm closing it out.

PR #249, opened 2017-01-15
Schedules Direct improvements.

enternoescape:


java/sage/Person.java
```diff
@@ -196,6 +235,14 @@ public boolean isPersonIdentical(Person testMe)
       for (int i = 0; i < awardNames.length; i++)
         if (awardNames[i] != testMe.awardNames[i])
           return false;
+      for (int i = 0; i < headshotUrls.length; i++)
+      {
+        if (headshotUrls[i].length != testMe.headshotUrls[i].length)
+          return false;
+        for (int j = 0; j < headshotUrls[i].length; j++)
+          if (headshotUrls[i][j] != testMe.headshotUrls[i][j])
```
Narflex: You're going to want to do an Array.equals comparison here. I got away with straight == in the above ones because I'm comparing Stringer objects which will then match because they are looked up from the DB table.

java/sage/Wizard.java
```diff
@@ -3115,13 +3116,10 @@ public boolean addSeriesInfoPublic(int legacySeriesID, String title, String netw
     return addSeriesInfo(legacySeriesID, title, network, description, history, premiereDate, finaleDate, airDOW, airHrMin, imageURL, people, characters) != null;
   }
   public SeriesInfo addSeriesInfo(int legacySeriesID, int showcardID, String title, String network, String description, String history, String premiereDate,
-      String finaleDate, String airDOW, String airHrMin, String[] people, String[] characters, byte[][] imageURLs)
+      String finaleDate, String airDOW, String airHrMin, Person[] people, String[] characters, byte[][] imageURLs)
```
Narflex: Please don't remove any of these 'add' methods...it'll break the other EPG plugin I'm using currently (which of course I'm going to ditch in a few months). Just overload it with another one. :)

java/sage/Wizard.java
```diff
@@ -4570,11 +4568,11 @@ public Show addShow(String title, String primeTitle, String episodeName, String
   }
 
   public Show addShow(String title, String episodeName, String desc, long duration, String[] categories,
-                      String[] people, byte[] roles, String rated, String[] expandedRatings,
+                      Person[] people, byte[] roles, String rated, String[] expandedRatings,
```
Narflex: Same thing as above, don't remove any of the add methods.

java/sage/Wizard.java
```diff
@@ -5473,12 +5579,16 @@ void removeWatched(Watched removeMe)
     t.remove(removeMe, true);
   }
 
-  public Person addPerson(String name, int extID, int dob, int dod, String birthPlace, short[] yearList, String[] awardList, short headshotImageID)
+  public Person addPerson(String name, int extID, int dob, int dod, String birthPlace, short[] yearList, String[] awardList, byte[][] headshotUrls, int mediaMask)
```
Narflex: Again, don't remove add methods.

java/sage/Person.java
enternoescape: I'm confused here. Are we not testing for the array to be identical to the other array? If I change != to == then we are saying that the objects are identical then their images are different.

enternoescape: If you're talking about using equals(Object), that's a 2D byte array, so there's not reason to go that route.

java/sage/Wizard.java
enternoescape: I understand, at last in the case, I'm fairly certain I put that one there, but I will fix that and note for future reference.

java/sage/Person.java
Narflex: My bad....I didn't see the other for loop in there. This is fine (I thought you were comparing the two arrays with ==...but that is not what you are doing).

Narflex:
And I didn't realize the Wizard methods you were modifying were ones you added...in that case I would have been fine with you leaving them there...but no reason to change it again, I'll get this merged. :)  Although at some point later we should probably change most of these horribly overloaded methods with may too many arguments to use some kind of builder or pass in the arguments via a structure. That's nobody's fault but my own it got this way. :)

enternoescape:
No worries. It's better to raise a flag and be wrong than merge buggy code. :)

PR #251, opened 2017-01-19
Added New API Methods for In Progress Sport Events.

enternoescape:
Closes issue #250.


Narflex:
All looks fine to me, I'll just wait until the Travis check finishes and then merge it (which'll be tomorrow morning).

enternoescape:
I wasn't completely sure if two of those API methods belonged in Show, but I couldn't think of a better categorization. I figured you'd let me know if you had a better suggestion.

Narflex:
I think the API categories you have them in is fine. Show makes the most sense because they are externalID based queries and that's a unique key for Show objects.

PR #253, opened 2017-01-27
Added editorials based on recommendations from Schedules Direct.

enternoescape:


java/sage/epg/sd/SDRecommended.java
```diff
@@ -0,0 +1,470 @@
...
+        continue;
+      String programId = recommendation.getProgramId();
+      if (programId == null || programId.length() == 0)
+        continue;
+
+      // Don't recommend movies that we have already seen or don't exist in the Wizard.
+      if (programId.startsWith("MV"))
+      {
+        Show show;
+        if ((show = Wizard.getInstance().getShowForExternalID(programId)) == null || show.isWatched())
```
Narflex: This looks like it will skip anything with an ID that starts with 00 because this first lookup on the uncorrected program ID will always fail in that case.

java/sage/epg/sd/SDRecommended.java
```diff
@@ -0,0 +1,470 @@
...
+          alreadyExists++;
+          skip = true;
+          break;
+        }
+      }
+      if (skip)
+        continue;
+
+      // Check if we could't record this show if we wanted to.
+      Show shows[] = Wizard.getInstance().getShowsForExternalIDPrefix(programId);
```
Narflex: Since you're using the full program ID here...I don't think you need to use the prefix lookup variant...you can just use the normal lookup.

java/sage/epg/sd/SDRecommended.java
```diff
@@ -0,0 +1,470 @@
...
+      String newID = SDUtils.fromProgramToSageTV(programId);
+      if (newID.length() == 12)
+      {
+        Show moreShows[] = Wizard.getInstance().getShowsForExternalIDPrefix(newID);
```
Narflex: Same here about the prefix lookup call

java/sage/epg/sd/SDRecommended.java
enternoescape: It does the conversion next. And tries that. This works for in both cases.

java/sage/epg/sd/SDRipper.java
```diff
@@ -2231,6 +2240,37 @@ public void run()
...
+        {
+          if (Sage.DBG) System.out.println("SDEPG Got less than " + minRecommendations +
+            " recommendations, adding intelligent recordings");
+          editorials = SDRecommended.getUsable(SDRecommended.getRecommendations(true));
+        }
+        SDRecommended.reduceByWeight(editorials, editorialImportLimit);
+        for (SDEditorial editorial : editorials)
+        {
+          Airing airing = editorial.getAiring();
+          String startTime = Sage.dfStd(airing.getStartTime());
```
Narflex: Use the date format of yyyy-MM-dd for this. It's expecting that in Wizard when it does the 'maintenance' process which basically garbage collects the DB. It'll then automatically remove editorials that are then over a week old (i.e. the date is more than a week in the past). If you use the m/d/yyyy format which Sage.dfStd applies, then it'll fail the date parsing in Wizard and never end up removing any of your added editorials. This is around line 1316 in Wizard that this happens.

Narflex:
Just a few things. Looks good though. :)

java/sage/epg/sd/SDRecommended.java
enternoescape: I see what you're saying.

java/sage/epg/sd/SDRecommended.java
enternoescape: In this case I'm looking for any shows that would match that prefix. The recommendation isn't as likely to match anything because they always come in prefixed with SH.

enternoescape:
I ran the resultant suggestions across my wife and she found two new things that interested her, so that was enough for me to think it's not worthless. :)

java/sage/epg/sd/SDRecommended.java
enternoescape: When I tried it the way you're suggesting I got a lot fewer hits on shows that otherwise definitely had airings.

Narflex: The point I was making is that when you pass a full programID into getShowsForExternalIDPrefix...it's the same as if you called getShowForExternalID because that method doesn't turn it into a prefix. Below in your code, you turn them into prefixes and call the prefix method...so you should be applying the same logic up here (if you add some debug logging, I'd expect you to only get one result returned from these first prefix calls)

Narflex:
My wife has been a SageTV tester since before it was even called SageTV...wife's feedback is always very important. :D

java/sage/epg/sd/SDRecommended.java
enternoescape: Oh, I see. Thanks for the clarification.

CraziFuzzy:
My wife certainly let me know that Joseph broke her morning news this morning.. ;-)

enternoescape:
I found few other issues with images that I didn't notice before. I can't get an image for everything and sometimes for some reason only some programs get images one day and the next day they don't have anything. It's really annoying. The end result is that you have editorials with nice images that make perfect sense where they are, then you have other images derived from the show object by some logic in the STV that don't always look right.

PR #271, opened 2017-03-21
Some Schedules Direct missing station improvements.

enternoescape:
It turns out that sometimes some airings actually do end up being found using the method we added to the Wizard and this prevents the station from being invalidated and re-downloaded. I changed it to just check for an airing at this exact time and if it doesn't exist or is a No Data airing, we invalidate the station. This is also based more strongly on data we need now instead of data we have now.

I also added some methods I added for Seeker2 to create timed/manual recordings because I really didn't like the idea of creating an airing outside of the Wizard within Seeker. It just feels like the wrong place to do this. I'm not ready to commit the Seeker2 core changes yet since I'm still slowly working out a few small, but known issues.

### DIFF
--- a/java/sage/Wizard.java
+++ b/java/sage/Wizard.java
@@ -64,6 +64,7 @@ import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.net.Socket;
 import java.text.DateFormat;
+import java.text.DecimalFormat;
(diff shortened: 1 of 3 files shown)


enternoescape:
I also fixed a null pointer exception that results from the fact that we sometimes have aliases without a normal person associated.

PR #272, opened 2017-03-28
Windows related build changes

jusjoken:
- use single version.h file for all windows exe's
- updated gradle to use git on windows to get build number
- installer updated to better handle version numbering
- service control now has sage icon and version info for consistency


enternoescape:
@jusjoken I'm wondering why we are using a shell to launch git to get the build version. I see 'sh -c' being used to invoke git which can be installed on Windows as a normal executable. I'm mostly asking because that adds a prerequisite that on the surface seems unnecessary.

jusjoken:
I got tired of manually updating the constants file so i found an example that worked. If there is a cleaner way in gradle to do it then feel free to update it as i am not familiar at all with gradle...yet.

I also now have a powershell script that fully builds native windows, creates the installer and uploads to bintray... one more small change to add the jar build and i will share that too.

enternoescape:
You're 99% there. This is just a matter of how git.exe is executed. I just wanted to know if there was a special reason why you went that route. It sounds like there isn't which is fine, I just wanted to know before I suggest some changes that might break something that you're accounting for.

Thanks for all of the hard work you have put into the Windows side of things. You're one of the few.

Issue #154, opened 2016-07-03
API call that allows a favorite to be disabled.


49studebaker:
There is a plugin that allows a user to disable a favorite, but the favorite's additional airing are hidden when the favorite is disabled. I would like to be able to view a favorite's additional airing when a favorite is disabled.

49studebaker:
I downloaded the latest SageTV download and it is now possible to disable a favorite without installing a plugin. I am still not able to view a disabled favorites "Aditional Airings".

skiingwiz:
I have most of the work done for this. I just need to find some time to finish it. I am adding a new API method to retrieve a favorite's airings ignoring the disabled state. The STV will then use the new method to show additional airings.

skiingwiz:
This is now available in 9.0.15. If you use the installer, you'll have to wait for a new version of that to be built. 

Someone with permissions may close this issue.

PR #274, opened 2017-04-05
Create new API, GetPotentialFavoriteAirings

skiingwiz:
Creates a new API, GetPotentialFavoriteAirings, that works exactly the same as GetFavoriteAirings, except that it returns the airings as if the airing were enabled, even if the favorite is disabled.

This will be used in an update to the STV to show upcoming airings for a favorite even when it is disabled.  (Issue #154)

### DIFF
--- a/java/sage/Agent.java
+++ b/java/sage/Agent.java
@@ -442,12 +442,12 @@ public class Agent extends DBObject implements Favorite
   }
 
   public synchronized Airing[] getRelatedAirings(DBObject[] allAirs, boolean mainCache, boolean controlCPUUsage,
-          boolean ignoreDisabled, StringBuffer sbCache)
+          boolean ignoreDisabledFlag, StringBuffer sbCache)
   {
     if (allAirs == null) return Pooler.EMPTY_AIRING_ARRAY;
 
     //Short circuit this if the agent is disabled (and we don't want to ignore that fact)
-    if(!ignoreDisabled && testAgentFlag(DISABLED_FLAG)) return Pooler.EMPTY_AIRING_ARRAY;
+    if(!ignoreDisabledFlag && testAgentFlag(DISABLED_FLAG)) return Pooler.EMPTY_AIRING_ARRAY;
 
     if (ENABLE_AGENT_AIRING_CACHE)
     {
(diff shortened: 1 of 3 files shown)


java/sage/Agent.java
```diff
@@ -465,7 +474,7 @@ void unbully(int weakerID)
       for (Show show : shows) {
         Airing[] airings = wiz.getAirings(show, 0);
         for(Airing a : airings) {
-          if (followsTrend(a, true, sbCache, true))
+          if (followsTrend(a, true, sbCache, true, true))
```
Narflex: This should pass through ignoreDisabled as the last value (I understand it'll have the same functionality, but it just looks more correct).

java/sage/Agent.java
```diff
@@ -474,7 +483,7 @@ void unbully(int weakerID)
       {
         Airing a = (Airing) allAirs[i];
         if (a == null) continue;
-        if (followsTrend(a, true, sbCache))
+        if (followsTrend(a, true, sbCache, false, true))
```
Narflex: This should pass through ignoreDisabled as the last value (I understand it'll have the same functionality, but it just looks more correct).

java/sage/api/FavoriteAPI.java
```diff
@@ -1176,6 +1179,23 @@ public Object runSafely(Catbert.FastStack stack) throws Exception{
...
       }});
+    rft.put(new PredefinedJEPFunction("Favorite", "GetPotentialFavoriteAirings", new String[] { "Favorite" }, true)
+    {
+      /**
+       * Returns a list of all of the Airings in the database that match this Favorite.  If the favorite is disabled
+       * this API call will ignore that fact and return the Airings that would match this favorite if it were not
+       * disabled.
+       * @param Favorite the Favorite object
+       * @return the list of Airings in the DB that match this Favorite
+       *
```
Narflex: Can you add an @since annotation here so we know what version this landed on? Also, go ahead and increment the micro version in Version.java as well and then use that new value here.

skiingwiz:
@Narflex  I have addressed all of your comments in the most recent commit.

Narflex:
Thanks!

PR #283, opened 2017-05-10
Disable hwaccels in libMpeg2Transcoder/ffmpeg

barberio:
This fixes a compile time error, caused by hwaccels in ffmpeg being dependant on h264 encoders. Since hwaccels aren't being used, they should be turned off along with all the other unused parts anyway.

### DIFF
--- a/build/buildso.sh
+++ b/build/buildso.sh
@@ -81,7 +81,7 @@ cd ../../third_party/ffmpeg
 make clean
 ./configure --build-suffix=-minimal --disable-static --enable-shared \
   --disable-decoders --disable-encoders --disable-parsers --disable-filters \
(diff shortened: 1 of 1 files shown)


Narflex:
What was the compile time error you were getting? No one else has run into this issue yet and many people have compiled it without issue.

barberio:
> /home/barberio/Developer/sagetv/third_party/ffmpeg/libavcodec/libavcodec-minimal.so: undefined reference to `ff_h264_get_slice_type'
> collect2: error: ld returned 1 exit status

Caused because --disable-decoders doesn't imply --disable-hwaccels, despite hwaccels having dependancies on various decoders. This will only show up if your build environment has the correct dev headers for ffmpeg's configure script to try and enable hwaccels. (ie, nvdec/libva...)

PR #288, opened 2017-05-22
Incremented version and added some Schedules Direct features.

enternoescape:


enternoescape:
That last commit should be a huge convenience for anyone that just wants to see what the alternate logos would look like. I've generally noticed they are brighter colored and work well with dark backgrounds. I realize that I have create a small race condition between extractGuide() and getUpdateChannelLogos() in that either one could be updating the channels at the same time and possibly cancelling the other one out. This is not a problem because normal users will not be fiddling with the logo selection frantically and if there is a small conflict of channel information due to toggling this within the first 500ms of extractGuide(), it will be resolved on the next update.

enternoescape:
I re-thought that and forgot that someone could be toggling the logos between two different clients and that might create a small problem. Synchronized it is!

enternoescape:
These commits address a few things I've had on the back burner. I want to keep Schedules Direct code current with whatever new things they add that we can actually use. This upcoming weekend might be a good opportunity to release a new version of SageTV.

PR #289, opened 2017-05-22
Fixed a problem with awk parsing in Ubuntu.

enternoescape:


build/serverfiles/startsagecore
```diff
@@ -31,9 +31,11 @@ if [ -e sagesettings ] ; then
 fi
 
 # If we are running Java 8 or 9, enable string deduplication. This has very measurable benefits in
-# multi-miniclient situations.
-java_version=$(java -version 2>&1 | awk -F '"' '/version/ {print substr ($2, 0, 3)}')
-if [[ "$java_version" == "1.8" || "$java_version" == "1.9" ]] ; then
+# multi-miniclient situations. There are some awk compatibility problems between Linux
+# distributions. We get an extra decimal in some and in others we get exactly the string we want.
+# Because of this, we check for both cases.
+java_version=$(java -version 2>&1 | awk -F '"' '/version/ {print substr ($2, 0, 4)}')
```
stuckless: I think this is fine, for now... but eventually we should probably change it to match this StackOverflow answer, and the modify the if statement to be -ge 8 to cover java 8, 9, 10, etc.  https://stackoverflow.com/a/40407601/614231

enternoescape: I like that a lot better. I'll put in another pull for that better approach later.  The reason I limited it to 8 and 9 is because we don't know if when 10 comes out if this is still going to be a valid option.

PR #295, opened 2017-05-31
A missing space in an if test prevents the Linux SageTV start script from running.

enternoescape:
Let's all agree to leave well enough alone on this one. It's working. We shall not change this piece of code ever again. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## Version 9.1.3 (2017-05-30)
+* Fix: A missing space in an if test causes the Linux start script to fail.
+
 ## Version 9.1.2 (2017-05-30)
 * Fix: Changed awk parsing to use sed to clean up the Java version check.
 * Fix: API methods GetFavoriteAirings() and GetPotentialFavoriteAirings() were returning all airings for keyword favorites.
--- a/build/serverfiles/startsagecore
+++ b/build/serverfiles/startsagecore
@@ -33,7 +33,7 @@ fi
 # If we are running Java 8 or 9, enable string deduplication. This has very measurable benefits in
 # multi-miniclient situations.
 JAVA_VERSION=$(java -version 2>&1 | grep -i version | sed 's/.*version ".*\.\(.*\)\..*"/\1/; 1q')
-if [[ "$JAVA_VERSION" == "8" || "$JAVA_VERSION" == "9"]] ; then
+if [[ "$JAVA_VERSION" == "8" || "$JAVA_VERSION" == "9" ]] ; then
     JAVAOPTS="$JAVAOPTS -XX:+UseG1GC -XX:+UseStringDeduplication"
 fi
 
--- a/java/sage/Version.java
+++ b/java/sage/Version.java
@@ -23,7 +23,7 @@ public class Version
 {
   public static final byte MAJOR_VERSION = 9;
   public static final byte MINOR_VERSION = 1;
-  public static final byte MICRO_VERSION = 2;
+  public static final byte MICRO_VERSION = 3;
 
   public static final String VERSION = MAJOR_VERSION + "." + MINOR_VERSION + "." + MICRO_VERSION + "." + SageConstants.BUILD_VERSION;
 


CraziFuzzy:
HA!.. picked a hell of a time to update unRAID, only to have the wife complain an hour later "The client's not seeing the server!"  Sure enough, it wasn't running.. ;-)

CraziFuzzy:
Still feel we need to somehow get the updates to come via docker container updates, not restarting the container.  Would prevent unintended updates like this causing mass problems, because it would take deliberate user intervention to update sagetv.

enternoescape:
That's easy. We would just hard code the SageTV version into the container. Then the container would need to be updated any time a new version was released. That would mean more work for @stuckless since he would then need to either automate the version incrementing of his containers or update them manually with each release.

enternoescape:
It does look like Travis CI is having some performance issues with our specific build type. They have some messages from 5/30 indicating that they are working with Google to sort out what's going wrong.

CraziFuzzy:
Yeah, hard coding the version into the script would be possible I suppose - would be a lot easier if the docker stuff was integrated into the gradle build, I'm guessing.  Part of a bigger job I suppose, since we need to see how to get the windows builds going as well.

enternoescape:
That's mostly what I meant by automating the hard coding. Something could be added to the Gradle script that would also update the Docker container so people know a newer version of SageTV is available. The problem SageTV has is that the software binary is mostly external to the container because the configuration is mixed in and because of the way it's written out, hard links wouldn't get the job done.

Alternatively there could a script to update the Docker container repo that when run checks if there's a newer version of SageTV and if so it pushes an update to the repo to use the new version. The "current" version could just be stored in a file that's checked on startup and that would make it very easy to update at the repo level by script. This approach would be more modular and not require any changes on the SageTV build side of things.

enternoescape:
For the record, I fully support what you're suggesting. I like the automatic upgrades, but I also can see why some people might be caught off-guard by them. We have a property that lets you go between versions easily, but some kind of prompt that you will be upgrading to a new version would likely be very helpful to most people.

CraziFuzzy:
I'm really wondering though just how hard it would be to switch the sagetv home folder separate from the binaries.  Just about all the files in question are generated on first runtime (wiz.bin, SageTV.properties, etc).  Might be worth an issue to track that improvement, even if it's just breaking apart some path variables at this point.  We shouldn't hold to legacy forever, if we have the option to move away from it.

enternoescape:
We had this discussion in the forums. I would be willing to put in the work if I didn't think it would break a lot of plugins along the way. You saw what happened when we just changed a classpath. This kind of change would require a shift in how plugin developers store persistent data. We might need to somehow shim the legacy plugins so they use the right locations and have some new requirements for new/updated plugins. The STV's would also need to be moved around which would likely cause it's own huge problem. I'm just not sure it's reasonable to attempt this.

I've been contemplating a plugin 2.0 system that provides the plugins with a folder they can do what they will with to ensure they place persistent data in a safe location that no other plugin will try to disturb. I was also thinking about ways to isolate plugins from the main JVM so that they can be accessed via the STV and they have full access to the main JVM through the API, but not introduce new dependencies to it or conflict with other plugin requirements. This would of course be through some kind of native bridge or local network connection and the plugins would likely run in their own JVM's. This would also keep any plugin from hosing the core program and might even make plugin development easier since you could then in theory run your plugin in the IDE with SageTV using it like it was installed the traditional way.

CraziFuzzy:
Like I said, it probably should have it's own issue topic so we can continue the discussion, because i do think it should be a long term goal.  I do have some ideas on how to deal with it, and some groundwork is already in the core to deal with it (the Sage.getPath function).  There are things in the core, however, that do not use this, and instead use System.getProperty("user.dir") instead that would need to be changed.  I'm sure there are plenty of plugins that also use user.dir.  The issue with the plugins COULD be handled by launching from the data directory, making that the "user.dir" as far as java is concerned, then any plugins that are using user.dir would be homed there, and newer/proper plugins would be geting the correct path from Sage.getPath.

enternoescape:
I missed where you said this should be it's own topic, but I agree. I can see how this could be in incremental moving forward goal.

PR #296, opened 2017-05-31
Fix bash syntax error in startsagecore.

leizleiz:
Double end brackets need a space in front.


enternoescape:
These are the same changes as pull request #295 .

PR #329, opened 2017-07-01
Removed LITE

stuckless:
Part of SageConstants cleanup.  Removed LITE usage.


Narflex:
That one was much easier to remove. :)

PR #335, opened 2017-07-11
Update makefiles to fix TARGET support for crosscompiling

jfthibert:


stuckless:
Have you tested this to cross compile?  And if so, do you have instructions on how to setup the cross compiling? And which architectures have you cross-compiled (Windows? Arm?)

jfthibert:
Yes, I have an additional change that I still need to cleanup which allowed me to build the server side for the raspberry pi 2+ (armhf). We used to also support building for mipsel a long time ago. I'll add instructions once I send that for review, the main idea is to pass TARGET=arm-linux-gnueabihf and have a properly setup cross compiler with at least zlib and freetype.

stuckless:
That is very cool.  I have a pi3 and I was doing some cross compiling (for another unrelated project) this past week and it got me thinking about sagetv and compiling for pi, so great timing :)  If we can get this working on a PI, then I can likely update the Travis scripts to create/deploy ARM builds as well that people can test out.

enternoescape:
This is really interesting. I too have a few RPi's that could be put to use.

wayner9:
I have several Pis as well so I would be happy to test things out.  I have everything from Pi3s to Pi Zero Ws to the older original Pi and Pi2.  There is also the Rock64 which is very similar to the Pi3 but a bit faster at 1.4GHz vs 1.2 GHz and supports 4K@60fps, has USB 3.0 ports, IR and a Gb ethernet port.  http://www.dailytech.com/Comparison+Rock64+vs+Raspberry+Pi+3/article42443.htm

CraziFuzzy:
Building for embedded/slim hardware would really be great if/when sage gets modularized a bit more, so not only can it be built, but maybe slimmed down as necessary to perform specific tasks suited to the limited hardware.

jfthibert:
I had to do a few steps manually and haven't had time to finish cleaning up the build scripts to work with cross compiling.
If you want to try the one I had built it's at [https://mu3d.com/sagetv/sagetv-server_rpi2.tar.gz](https://mu3d.com/sagetv/sagetv-server_rpi2.tar.gz), it's been running fine since the weekend.

The main setup steps are installing the lite version of the latest raspbian, tweaking the usb power settings (using a good PS), adding a usb hard drive and tuner, installing java then extracting the tar.gz to /opt/sagetv/server and running startsage.

PR #336, opened 2017-07-11
Removed COMPACT_DB flag

stuckless:
As per Jeff's suggestion, removed the COMPACT_DB flag.


Narflex:
I'm assuming you also did a test load/save of the DB to make sure it didn't break anything there. :)

stuckless:
Well, I never actually loaded it :)  But i promise I'll do some testing before we do a new SageTV release.

Issue #338, opened 2017-07-12
Problems with OTA HDHomerun Scan


texneus:
I am having an issue getting a channel scan to function correctly using an HDHR connected to OTA.

Server and Placeshifter are running on Kubuntu 16.10, although it does not seem critical (I have also tried various combinations of the server on Ubuntu Server 16.04 LTS and the SageTV 9 Windows client with no change in results).  The only place where I can get a completely successful channel scan is using the Windows server with the Windows client.  I have migrated to Server and Placeshifter on Kubuntu 16.10 on another machine so I can continue to upgrade my SageTV 9 system hardware & software.

HDHR is the original, older (white) model, apparently now known as the HDHR Discover (this is what it reported when I updated the firmware).  Firmware is the latest from Silicon Dust's website as of last weekend (7/8/17) (I updated from 2014 firmware with no effect on this problem).

The problem is upon completion of the channel scan I have several clearly bogus channels simply numbered sequentially, starting with "-", then 2, 3, 4, ... up to 22 but sometimes up to 23 or 24.  After these channels, the expected results are shown and they appear complete.  I can simply disable the channels and move on, but I suspect this is aggravating other issues in the docker version of the server (where I really want to be) that make channel scan results basically useless for me.

In the log I see several entries that resemble the following:
```
Tue 7/11 23:17:27.863 [AsyncChannelSet@17c02a1d] ---  Processing scan result for digital tuner physical channel number 44
Tue 7/11 23:17:27.863 [AsyncChannelSet@17c02a1d] -----  scan result: 

Tue 7/11 23:17:27.863 [AsyncChannelSet@17c02a1d] Repeat channel name: [
]; adding a number to the channel name: [
19]
```

So it appears certain channel names are not being passed or processed correctly from the HDHR.  Once again the Windows SageTV 9 server/client appear immune to this issue.  Likewise a channel scan using HDHomerun_config within Ubuntu provides as-expected results.

Attached is the log from SageTV server on Kubuntu 16.10 from this evening.  I don't see a corresponding placeshifter log?

[sagetv_0.txt](https://github.com/google/sagetv/files/1140897/sagetv_0.txt)

Other logs may also be found in the related thread in the SageTV forums:  https://forums.sagetv.com/forums/showthread.php?t=64720

CraziFuzzy:
[...]

PR #342, opened 2017-07-21
Fix: ATSC Scanning Blank and Garbled Numbers

CraziFuzzy:
Fix for scanning issues discussed in issue #338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * New: Add Schedules Direct lineup by ID.
 * Change: Removed ZZZ from Schedules Direct Regions because it doesn't do anything.
 * Fix: VOB and MP4 subtitles locking methods were not being called.
+* Fix: Fixes to HDHomeRun (and probably others) ATSC Scanning returning blank and garbled channels.
 
 ## Version 9.1.5 (2017-06-19)
 * Fix: Carny throws a null pointer exception if a show has a null title.
--- a/native/ax/Channel-2/Channel.c
+++ b/native/ax/Channel-2/Channel.c
@@ -2873,8 +2873,8 @@ int scanDVBTChannel( CHANNEL_DATA *Channel, char* tuningString, char** scanResul
 	//SageTVReleaseScanChannel( Channel->Dev );
 	if ( found ) 
 	{
-		pos += SPRINTF( channels_buf+pos, channels_size-pos, "\r\n" );
-//		saveScanTuningTable( Channel, "frq" );
+//		pos += SPRINTF( channels_buf+pos, channels_size-pos, "\r\n" );
+		saveScanTuningTable( Channel, "frq" );
 	}
 	SageLog(( _LOG_TRACE, 3, "DVB-T Scan Info:%s %d.\r\n", *scanResult, found ));
 
--- a/native/ax/Channel-2/Channel.c
+++ b/native/ax/Channel-2/Channel.c
@@ -2280,7 +2280,7 @@ int scanATSCChannel( CHANNEL_DATA *Channel, char* tuningString, char** scanResul
 	//SageTVReleaseScanChannel( Channel->Dev );
 	if ( found ) 
 	{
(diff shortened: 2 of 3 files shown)


Narflex:
Have you had a few people test this to make sure it doesn't break existing working systems as well as fix the one who has issues? Your analysis sounded pretty thorough, but testing is always good. ☺️

CraziFuzzy:
Only testing on my own system, where I was able to reproduce the OP's issue.  Figured these were some pretty 'safe' changes.

CraziFuzzy:
I never was able to determine the source of the odd characters getting into the channel name.  What was odd, and reproducible, was that any channel name that used the full 7 characters had the same garbled response following the 7, when run in Docker, but NOT occurring with the same build on bare ubuntu.  My only thoughts were perhaps the docker has a different default encoding set so that the non-initialized string was being interpreted differently there.  In any case, nulling it out first take away any question.

Narflex:
I'm on vacation right now, and I can review it in more detail when I get
back to see if I come to the same conclusion...or if a couple other people
test it I'm ok with merging it. The changes themselves aren't clearly safe
because as individual lines they appear to be changing how things
work....but I do understand that if I was to look at it overall I'd likey
come to that conclusion.

Jeff Kardatzke
Sent from my Android

On Jul 21, 2017 9:19 AM, "Christopher Piper" <notifications@github.com>
wrote:

Only testing on my own system, where I was able to reproduce the OP's
issue. Figured these were some pretty 'safe' changes.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/google/sagetv/pull/342#issuecomment-316924879>, or mute
the thread

CraziFuzzy:
No worries.  This is not a 'new' problem, so much as it's one that really didn't exist en masse until people started moving to Linux, and all it really causes is needing to massage the lineup a bit after the scan.

Not adding the line break to the scaninfo is probably the most significant change, though having spent a bit of time last year in the parsing code in the STV, I really have no idea why that was added in, as nothing indicates any reason for line breaks in the semi-colon delimited channel list.

In any case, enjoy the vacation.

texneus:
I'd be more than happy to test, but I'm going on vacation myself in a week, so it will need to be quick!  Let me know what I can do.

On a side note, I read today that Dallas-Ft. Worth is one of the first markets to go through a spectrum repack in preperation for ATSC 3.  Channels begin moving around this fall, and allegedy some of the low power stations I was getting are already shutting down (sub channels turned off, others have just color bars, probably as existing contracts run out).  The channel scanner will probably get a good workout the next few months...

CraziFuzzy:
Thanks @texneus .  I could send you the .so files zipped up and you could drop them into your sage folder to test, but frankly, I don't know if that's really necessary.  I have pretty much your exact setup, and confirmed the bug before, and the fix after.  What would be more helpful is users who do NOT have the same config (different tuners, different OS's, etc) to test it to make sure it didn't break them.

Narflex:
Fuzzy,

I took a look at this again now that I'm back and I really would like to have somebody test this on a Windows setup and then also somebody else who has an already working setup on Linux to test it. As you're aware, this code is really ugly and complex and it's possible there's some other dependency somewhere else on how this works that could get broken by this change.

jfthibert:
I took a quick look at this, according to

https://www.atsc.org/wp-content/uploads/2015/03/Program-System-Information-Protocol-for-Terrestrial-Broadcast-and-Cable.pdf

the VCT name is stored as 7 characters of UTF-16, the whole vct is already 0 initialized at the top of the loop and the name is defined as `char name[9*2];`. I'm not sure why it wasn't set as 8*2 but the proper fix in the loop is to simply copy the first 14 bytes. The current proposed fix seems right after removing the unnecessary memset.

CraziFuzzy:
@jfthibert that's funny, i stared at this code so many times, and never even noticed that it is already nulled out.  I was actually unfamiliar with what declaring with ={0} did in the case of struct's.  I can certainly remove the unnecessary memset - as you said, the core of the issue was grabbing more than the 14 bytes of the name.

Issue #347, opened 2017-08-08
Review Garbage Collector options and settings


JREkiwi:
Just throwing some thoughts out about GC as I don't have test environments to evaluate.

Following the change to G1GC in 9.1 there appears to been an increase in reported system crashes relating to OOM errors and initiating Native threads.  These all appear to be occurring in Windows systems (32bit JVM)

From my reading, it seems that G1GC is better suited to larger heap sizes and CMS is probably better suited for the limited heap size available in 32bit JVM.  Additionally Java 8 introduced MetaSpace which is stored in Native Memory and by default has no size limit and G1GC currently seems to have a number of bugs with managing MetaSpace, which may explain why disabling the Optimized Image Loader in SageTV reduces the frequency of crashes, although that could be a different issue altogether.

Some possible settings for testing of the CMS GC for 32bit JVMs
`-XX:+UseConcMarkSweepGC -XX:+CMSIncrementalMode -XX:+CMSIncrementalPacing -XX:+ParallelRefProcEnabled -XX:+CMSScavengeBeforeRemark -XX:+CMSClassUnloadingEnabled -XX:+ExplicitGCInvokesConcurrentAndUnloadsClasses -XX:InitiatingHeapOccupancyPercent=70`

Additionally there seems to be some additional tuning parameters that could be tested for the G1GC 
`-XX:+UseG1GC -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled -XX:+ExplicitGCInvokesConcurrent -XX:MaxGCPauseMillis=200`

On my Windows system the G1GC takes 2secs to run, while the default Java 8 GC takes 400ms.

Currently I am running Java 8.144, SageTV 9.1.5, but using the 9.0 service executables to start SageTV with the default GC.

John

enternoescape:
2 seconds? That's crazy. On my test VM I'm seeing less than 40ms. I just submitted a pull to change it back to the default for Java 8, so the next release should be good for everyone again.

PR #349, opened 2017-08-09
Removed use of G1GC in Windows due to possible memory leak issues.

enternoescape:


enternoescape:
Issues #347 explains some of the rational behind this pull.

PR #353, opened 2017-08-13
Correction as the icon file was named incorrectly

jusjoken:


JustFred51:
Confirmed .. building the entire Solution from within Visual Studio is now working.  
Thanks Ken

PR #362, opened 2017-09-04
Changes to use special writeUTF and readUTF to handle strings larger …

jusjoken:
…than 64k. Fixes GetFileAsString api call if file is larger than 64k

### DIFF
--- a/java/sage/SageTVConnection.java
+++ b/java/sage/SageTVConnection.java
@@ -185,213 +185,7 @@ public class SageTVConnection implements Runnable, Wizard.XctSyncClient, Carny.P
     underlyingOutStream = new java.io.BufferedOutputStream(mySock.getOutputStream());
     underlyingInStream = new java.io.BufferedInputStream(mySock.getInputStream());
     outStream = new MyDataOutput(underlyingOutStream);
-    inStream = new java.io.DataInput()
-    {
-      public final int read(byte b[]) throws java.io.IOException
-      { return underlyingInStream.read(b, 0, b.length); }
-
-      public final int read(byte b[], int off, int len) throws java.io.IOException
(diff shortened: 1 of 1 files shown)


Narflex:
I think a cleaner way to fix this that avoids duplicating more of the writeUTF/readUTF logic is to just use the MyDataOutput class that already exists in this file where the writing is done in place of the 'new java.io.DataOutputStream' calls. And then take the anonymous class at line 188 and make that MyDataInput and then use that in place of all the 'new java.io.DataInputStream calls'.

Or maybe there's some reason why I didn't do that before....but if there is I can't remember why right now. :)  Let me know what you think.

jusjoken:
I thought so too but MyDataOutput and the code at 188 seem to change a lot more than just write/readUTF and I have NO familiarity with this type of byte code so I was not sure if it may break more than I was aiming to fix.  I could give it a try if you think that the risk is worth it as I am not sure what all uses this code nor what tests to perform to validate I did not break other things.
Thoughts?

Narflex:
I just compared those two classes in the SageTVConnection class to what is in the JDK source code and they're doing the exact same thing except for the modified techniques we have for read/writeUTF. So I think it's totally safe to do what I was suggesting. :)  I know this code very well, so I'd be able to spot anything wrong in the code review (although, please run a test by executing it after you build it to make sure it does actually work fine and fixes the original problem). Sound good?

jusjoken:
Sounds good.  I will take another shot at it and test then post the results.
Thanks

jusjoken:
I have tested this with a windows client connected to a windows server as well as a linux server as well as tested the specific case in Gemstone that was failing previously with larger than 64k files and all those tests pass.

PR #373, opened 2018-03-08
update libhdhomerun to 20170930

jason-bean:
It's been a few years since libhdhomerun was updated.


Narflex:
Can you also bump the minor version number for SageTV in this commit so we can track if people have issues relating to this update?

PR #386, opened 2018-10-09
Fix wiz.bin DB crashes when a description is longer than 32KB

seansd:
Don't write to Show desc data via bytes when the length is greater than 32KB, instead opting to use the desc string.  I am creating the PR to create the discussion. . .I'm pretty sure there should be something done on the read side as well around line 477, b/c it's calling readShort to get the size and I'm not sure that's good either, but not sure how to fix that. . .

### DIFF
--- a/java/sage/Show.java
+++ b/java/sage/Show.java
@@ -675,7 +675,7 @@ public final class Show extends DBObject
       out.writeShort(barr.length);
       out.write(barr);
     } else {
-        out.writeUTF(descStr);
+      out.writeUTF(descStr);
     }
 
     out.writeInt((categories.length == 0) ? 0 : (useLookupIdx ? categories[0].lookupIdx : categories[0].id));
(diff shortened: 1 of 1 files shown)


seansd:
I signed the CLA, sean.sd@gmail.com

Narflex:
I'm totally baffled as to how this wasn't fixed before...I have various fixes in other places for handling large strings and I *thought* the original reason was because of large descriptions in the Show objects on imported MediaFiles...but anyways, clearly this part wasn't done correctly. The best way to fix this I think would be to just get rid of those byte arrays it's using and instead always read/write the description as a UTF. It'll then be backwards compatible and support adding of any new objects that have a description larger than 32K.

seansd:
We were wondering on the forums why the byte arrays were used at all. . .so good ok.  I can try to address this in the PR as well but how should i tackle it?  Just use descBytes only IFF there's no descStr? Or just get rid of descBytes completely?

seansd:
Line 472 is what concerns me. . . as I said, but assuming I can just use descStr always then I don't think you even need to do a readShort / readFully but not sure where descStr would come from then

seansd:
Also here's the link to the forums where we were discussing the issue: https://forums.sagetv.com/forums/showthread.php?t=65692

Narflex:
I'm suggesting getting rid of descBytes completely and also episodeNameBytes as well while you're at it. They were there originally as a memory usage optimization since Java uses 16 bit chars...and most chars used for this data are within 8 bits....so these cut down on memory usage a bunch. Then just use read/writeUTF with the descStr and episodeNameStr vars and that should be good.

seansd:
:+1:

seansd:
@Narflex have a look, changes done, I left the the deprecated bytes properties for now just b/c I was worried of other unforseen side effects, but if you want me to remove them i can

java/sage/Show.java
```diff
@@ -351,34 +320,35 @@ void setDontLike(boolean x)
...
-    rated = fromMe.rated;
-    ers = fromMe.ers;
-    year = fromMe.year;
-    pr = fromMe.pr;
-    bonuses = fromMe.bonuses;
-    language = fromMe.language;
-    originalAirDate = fromMe.originalAirDate;
-    seasonNum = fromMe.seasonNum;
-    episodeNum = fromMe.episodeNum;
+
```
Narflex: Please undo all of these alignment formatting changes, that's not fitting with the style of all the other code in the SageTV codebase.

Narflex: But do of course remove the lines that use episodeNameBytes of descBytes.

java/sage/Show.java
```diff
@@ -456,38 +426,30 @@ public boolean isIdentical(Show other)
...
-      in.readFully(episodeNameBytes);
-    }
-    size = in.readShort();
-    if (size == 0)
-      descStr = "";
-    else {
-      descBytes = new byte[size];
-      in.readFully(descBytes);
-    }
+    int size        = 0;
```
Narflex: Please don't reorganize code that's not part of the change or modify formatting in code that's not part of the change...makes it harder to review and isolate faults that come from changes.

java/sage/Show.java
```diff
@@ -1536,9 +1502,9 @@ public static String getImageIdUrl(int packedImageId, int rootId, int altFilmId,
   long duration;
   Stringer title;
   volatile String episodeNameStr;
-  byte[] episodeNameBytes;
+  byte[] episodeNameBytes; /* deprecated no longer used */
```
Narflex: Remove both of these so we know for sure they are no longer used...because if they are it won't compile. :)

seansd:
OK i'm going to submit a new PR and close this one so that I can get it cleaner and avoid formatting changes

PR #387, opened 2018-10-10
Fix long desc / episode name crash 

seansd:
When a show desc or episdode name is longer than 32KB the import will fail due to reading old byte arrays and their lengths as shorts.  Remove the byte arrays in favor of just using strings so the application doesn't crash.

### DIFF
--- a/java/sage/Show.java
+++ b/java/sage/Show.java
@@ -163,43 +163,11 @@ public final class Show extends DBObject
 
   public String getEpisodeName()
   {
-    if (episodeNameBytes != null)
-    {
-      // Doing it this way should be thread safe, we may create the String twice but we'd always
-      // do it from a valid byte array
-      byte[] testBytes = episodeNameBytes;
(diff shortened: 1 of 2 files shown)


Narflex:
I'll merge it once Travis builds it and gives the green light. You *may* need to change your GIT settings to use the email address you signed the Google CLA with...I'm not sure whether or not it'll let me merge your change otherwise. I did look it up internally based on your email and see you did sign...so I'll try to merge...but it may block me...we shall see. :)

seansd:
@Narflex do you have any idea when a new release version / installer would pick this up, i assume you have other stuff in the pipe for said release

Narflex:
I haven't built a new release in a long time...other people handle that now. :) I'm pretty sure the Windows ones are built manually...but the Linux ones I think are built automatically.  For this fix, you just need to replace the Sage.jar file so you should be able to do that yourself pretty easily.

Issue #389, opened 2018-10-29
Sort by Orig Airing Date violates its contract


wnjj:
User gary201 was having an issue where his recordings list would sometimes be completely blank.  Log files revealed an exception in the sorting routine where it "violates its general contract."  After debugging, I concluded the sort routing had a bug where it could violate the contract.  Details are in this post to his thread: https://forums.sagetv.com/forums/showthread.php?p=616949&postcount=20.

I built a debug jar file with changes to the sort that has been running clean on his system for a month when it used to fail within days.  I will submit a pull request with a change soon.

PR #390, opened 2018-10-29
Fix Issue #389, SortByOrigAiringDate violating its general contract

wnjj:
https://github.com/google/sagetv/issues/389

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,3 +26,4 @@ Martin Weber Nissle <webernissle@gmail.com>
 Chris Danser <chris.danser@gmail.com>
 Ken Birch <usjusjoken@gmail.com>
 Joseph Shuttlesworth <rightchicken@gmail.com>
+Warren Juenemann <warren.juenemann@gmail.com>
--- a/java/sage/api/Database.java
+++ b/java/sage/api/Database.java
@@ -1619,14 +1619,16 @@ public class Database {
             {
               int n1 = s1.getSeasonNumber();
               int n2 = s2.getSeasonNumber();
-              int x = n1 - n2;
-              if (x != 0 && n1 != 0 && n2 != 0)
-                return x;
+              if (n1 > n2)  // larger season is older/earlier show
+                return -1;
+              else if (n1 < n2)
+                return 1;
               n1 = s1.getEpisodeNumber();
               n2 = s2.getEpisodeNumber();
-              x = n1 - n2;
-              if (x != 0 && n1 != 0 && n2 != 0)
-                return x;
+              if (n1 < n2)
+                return -1;
+              else if (n1 > n2)
+                return 1;
             }
             Airing a1 = getAirObj(o1);
             Airing a2 = getAirObj(o2);
--- a/java/sage/api/Database.java
+++ b/java/sage/api/Database.java
@@ -1572,7 +1572,25 @@ public class Database {
             }
           }
           if (sortie != null)
-            java.util.Arrays.sort(currData, sortie);
+          {
+            try
+            {
+              java.util.Arrays.sort(currData, sortie);
+            }
+            catch (Exception e)
+            {
+              if (sortTech.toString().equalsIgnoreCase("getoriginalairingdate"))
+              {
+                // We know we violate the sorting contract in this method, so if it fails do it the safe way
+                sortie = optSortMap.get("safegetoriginalairingdate");
+                java.util.Arrays.sort(currData, sortie);
+              }
+              else
+              {
+                throw e;
+              }
+            }
+          }
           else
             java.util.Arrays.sort(currData);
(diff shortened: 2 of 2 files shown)


Narflex:
This is a tricky one...and I can tell you exactly why I had it the way it was. :)  I had two episodes of Survivor left for a season...the finale and the aftershow.  I hit Play on the group, knowing that SageTV would play them in order and then it started playing the aftershow FIRST and spoiled for me who was in the final 3 for that season.  The reason this happened was that the aftershow did not have a season/episode number attached to it...but the finale did. So because 0 is less than whatever season that was...it showed me them out of order.  They also had the same original airing date...which is why it also has the Airing time sort at the end.

When I made this change, I knew I was effectively breaking the sort contract...but there didn't seem to be an alternative to that unfortunately that I could think of.  It's clearly broken in some cases the way it is though...but maybe another ugly hack for catching the IllegalArgumentException and then using the newer variant in that case?  Spoilers suck..so those are usually my higher priority in preventing that from occurring (crashes are worse though).

wnjj:
Well that makes sense but it does present kind of an unsolvable problem then doesn't it?  For what it's worth, I never knew you could play a group in order like that.  I guess I've always just picked what I wanted.  Is there any way for the "play all" to get back some kind of flag if the sort order is compromised (i.e. missing fields like season/episode with same airing date) and refuse to play the list?

Too bad "original airing date" doesn't include the time to make it sufficient.

Narflex:
That's an interesting idea you subliminally put in my head. :)  One of the flaws of original airing date has been just that...it doesn't consider time.  HOWEVER, I think it would be totally valid to take the time from the Airing time and adjust the original airing date by that amount instead so you then effectively have an original airing time (you would only do this if the airing date matched the original airing date).

Having Play All try to detect this would be too weird.

It'd probably be best to do this in Wizard.addShow (it won't fix things from the past, but I think it'd be better to have this correct).  I'd then be fine with the other change you have here since it would solve the only case I'm aware of that's against it.

wnjj:
I'm a little unsure of how to do that in addShow.  From what I can see in the EPG ripper, Shows are added first, then Airings associated with the Shows are added next.  Doesn't that mean addAiring would need to adjust its corresponding Show?  Could it do this by comparing if the startTime is on the same date as the Show's originalAirDate?  Maybe using Sage.dfStd(originalAirDate) == Sage.dfStd(startTime) and setting originalAirDate = startTime if so?  Any help/code you can provide here would be great as I'd like to keep going on the 64-bit Windows work.  If you want to submit changes yourself, I'm even fine with cancelling this pull request or I can add whatever's needed to it.

Narflex:
That's a very valid point....you can't do what I said because Show doesn't know enough.  So this should only be done in the original airing date comparator then.  I'd just go with a trivial solution where if the original airing date is within 24 hours of the airing time, then use the airing time instead of the original air date and sort based on that.

This won't be perfect, for example if they show episodes 1 & 2 on a day, and then later that night repeat them and you recorded 2 earlier in the day and 1 later...then it'll have them inverted (whereas the current technique would sort them properly).

But this needs to be a sort that doesn't violate the sorting contract...and I see no real way to make it perfect....damnit...I hate these problems.  I think the super hacky way I mentioned before may be the best bet.

So do this:
Replace line 1575 with this code (needs indentation):
try {
  java.util.Arrays.sort(currData, sortie);
} catch (Exception e) {
  if (sortTech.toString().equalsLowerCase("getoriginalairingdate")) {
    // We know we violate the sorting contract in this method, so if it fails do it the safe way
    sortie = optSortMap.get("safegetoriginalairingdate");
  } else {
    throw e;
  }
}

Then clone the new version of the sorter to instead be put in the map with "safegetoriginalrigindate".

Sound good?  I'd do it myself but I have no spare time currently.

wnjj:
Sounds good.  I just needed a little help with the 'how' for the hacky method and you've provided that.

Narflex:
Please also let me know that you loaded up the JAR and tested that it works. :)

wnjj:
Sorry.  Was trying to do this from work and it keeps using the wrong username (failing the CLA).  I'll do this when I get home after I test the jar.  ;)

wnjj:
By "testing" you mean make sure the jar loads up and runs OK for a normal system, right?  I can't test the "fix" for the exception case with bad data since it wasn't ever happening on my system.  The code change will leave the non-exception path completely intact so I think making sure it works in the normal case should be sufficient to not break anyone else's system.

wnjj:
Also, did you mean to include another call to "java.util.Arrays.sort(currData, sortie);" again after setting sortie to the safe method?  I assume so.

Narflex:
Correct on both.

wnjj:
Jar ran fine on my system.

Narflex:
Thanks!

Issue #383, opened 2018-08-25
SeriesInfo.java NullPointerException


graywolfusmc:
SeriesInfo.java NullPointerException error is causing write to wiz.bin to fail, leaving Wiz.bin.tmp
Restarting clears this up but you lose anything done (recordings set, flag changes [watched, archived, etc.])

Can someone add code to prevent the NullPointerException from causing this issue and preventing the wiz.bin from getting written?

Thu 8/23 18:11:53.662 [EPG@11bd0e2] Wizard writing out table info for SeriesInfo
Thu 8/23 18:11:53.906 [EPG@11bd0e2] Exception during Maintenance:java.lang.NullPointerException
Thu 8/23 18:11:53.907 [EPG@11bd0e2] java.lang.NullPointerException
Thu 8/23 18:11:53.907 [EPG@11bd0e2] 	at sage.SeriesInfo.write(SeriesInfo.java:201)
Thu 8/23 18:11:53.907 [EPG@11bd0e2] 	at sage.Wizard.saveDBFile(Wizard.java:7930)
Thu 8/23 18:11:53.907 [EPG@11bd0e2] 	at sage.Wizard.maintenance(Wizard.java:2369)
Thu 8/23 18:11:53.907 [EPG@11bd0e2] 	at sage.EPG.run(EPG.java:645)
Thu 8/23 18:11:53.907 [EPG@11bd0e2] 	at java.lang.Thread.run(Unknown Source)
Thu 8/23 18:11:53.907 [EPG@11bd0e2] sage.epg.sd.SDRipper@df4c8 needs an update in 15:49:35
Thu 8/23 18:11:53.907 [EPG@11bd0e2] EPG needs an update in 949 minutes
Thu 8/23 18:11:53.907 [EPG@11bd0e2] EPG's works is done. Waiting...

Narflex:
Blocking that NPE may just then push the error to a different place. I'm not really sure how this could have happened in the first place...I'd also expect it to clean itself up after restarting since the database ensures all objects are 'valid' when it is loaded...and checking for nulls in the array causing the NPE is one of the validation steps.  Is this still a problem for you, or did it just happen this one time?

wnjj:
Hi Jeff,
I built a debug jar for him to try that simply blocks any null people from being added to the seriesinfo in the first place.  That way any other place that attempts to access it won't have an issue.  The code change (along with a debug println is in this post:  [https://forums.sagetv.com/forums/showthread.php?p=616487#post616487](url)

[...]

PR #391, opened 2018-11-02
Block NULL people/characters in addSeriesInfo.

wnjj:
This addresses issue #383 where NULL people were added to SeriesInfo, which in turn caused exceptions when accessed later.  User graywolf has been running a jar with this modified code for the past 3 weeks with good results, including a debug log event showing a caught NULL.  I recompiled and tested the jar fresh from this PR commit this evening.

I left the debug log statement in there because it should be rare and can help identify the source of these NULLs.  In this case, I believe BMT may be providing the seriesinfo data.


Narflex:
Sorry for the delay...was moving into a new house. :)

wnjj:
No worries.  Welcome home.  ;)

Issue #393, opened 2018-11-23
Debian packaging has missing and out of date dependencies


skiingwiz:
The debian packages have missing and out of date dependencies.  Both the [client](https://github.com/google/sagetv/blob/master/build/ubuntufiles/server/DEBIAN/control) and the [server](https://github.com/google/sagetv/blob/master/build/ubuntufiles/server/DEBIAN/control) list java5-runtime and java5-sdk. Since the minimum version of Java for the codebase is now Java 7, those should be updated to java7-runtime and java7-sdk.

In addition, there are some things that are definite dependencies that aren't listed.  The thing that caught my eye is net-tools, which provides ifconfig.  My laptop kubuntu didn't install that by default, which lead to my client reporting a MAC address of all zeros.

I'm happy to make and test the changes.  I've opened the issue to see if there are other missing dependencies that anyone knows about.

PR #394, opened 2018-12-01
Update deb dependencies

skiingwiz:
Resolves Issue #393 

I tested by installing.  installation was successful and installed net-tools.


Narflex:
Thanks!

PR #400, opened 2019-01-04
Windows 64-bit version

wnjj:
Primary changes are to support compiling a 64-bit Windows version.  A couple of other minor updates are included and listed in the changelog.


Narflex:
I would greatly prefer if this was broken up into multiple pull requests. It'll be much easier to isolate faults if this breaks something that way by building without a specific commit (and with this many changes, there's a very high chance that something else will get broken unrelated to 64-bit Windows where it was heavily tested).  It'll also make reviewing it much easier if it's broken up.

wnjj:
Ok.  I'll just cancel this one for now.

PR #403, opened 2019-01-08
MinGW native build.

wnjj:
These changes allow compiling of FFMPEG and its libraries on MinGW (MSYS2).  It also builds the pushreader lib for use in building the StrmDeMux filter plus the freetype, Mpeg2Transcoder and imageloader DLL's.


Narflex:
This also has a couple of other minor changes to cross-platform headers for faac, gif, x264, ffmpeg...so if new issues pop up in those on Linux, reverting this change would be worth trying to fix them. (all the changes look fine, I just wanted to document that)

PR #406, opened 2019-01-11
Change decoder/render defaults for new installations.

wnjj:
This change was requested in this thread:  https://forums.sagetv.com/forums/showthread.php?t=65769

The idea is to improve the initial user experience since V9 without going over V7 and 64-bit V9 do not have the licensed SageTV MPEG decoders.


Narflex:
There's possibly a better way to do this...in sage.DShowTVPlayer.autoOptimize(boolean).  That code path pretty much doesn't do anything anymore because of the default values for the preferences...it was mainly for fixing upgrades when we added our own decoders. But you could do more advanced things in there if you detect a brand new setup. It's designed to be something where you can intelligently select the best decoders/renderers depending upon what filters they have installed.

Take a look at that and let me know what you think.  It's definitely reasonable to approve the change you have here...but you could likely make it even more robust by updating that method to work better.

wnjj:
I looked at that code briefly and I think I understand what it's doing.  Please correct me if I'm wrong.  It checks for the "video_decoder_filter" and "audio_decoder_filter" properties and if length==0 (i.e. not defined), it will search through the installed filters trying to locate them.  At the end, it may end up picking "default" if none it knows about are found.  Finally, it writes the choice into the properties file.  The pull request code skips all of that, ending at the same place (i.e. "default" is already there).  Fixing it in the Java code would still allow for a V7->V9 32-bit install to pick up the SageTV MPEG decoders instead of making all users choose default.

So doesn't this code end up selecting default anyway if the SageTV MPEG decoders are not found?  If so, doesn't that make the suggested default properties for those filters unnecessary?
Would a good fix including searching for LAV before ending up at default?

As for the EVR, should the change be made to the code here instead? 
 https://github.com/google/sagetv/blob/master/java/sage/DShowMediaPlayer.java#L199-L211
Is EVR sometimes not available and therefore not a good default to use?  What logic should be used?

I'll admit, I'm still learning about all of these filter and render options and was taking others' lead here.

wnjj:
Reading further I found that "Overlay" is old, "VRM9" is XP and "EVR" is for Vista and newer.
This code selects the renderer:  
https://github.com/google/sagetv/blob/master/java/sage/Sage.java#L118

It looks like if VISTA_OS is set, EVR should end up being selected anyway when nothing is found in the properties file.  From what I can see on https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions, Windows 10 version is likely reported as 10.0 which wouldn't set this VISTA_OS flag.  Is the issue here that Win10 is not detected as "Vista" correctly?  If so, are there other consequences?

wnjj:
I tested out the use of autoOptimize with the empty decoder settings in the properties file and it found and setup the LAV filters by default.  EVR for normal & DVD on a Windows 7 machine also came up selected by default.  I fixed the log filter name typo for LAV in the 2nd commit.

The VISTA_OS detection was also tested including simulated cases with "10,0" returned for os.version since I don't have a Win10 machine.  This implementation should be future-proof.

CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # Change Log
 
-* New: Add EXEMultiTunerPlugin, HCWIRBlaster & USB-UUIRT VS projects (Windows)
-* 64-bit code and VS project updates (Windows)
-* Enabled MSYS2/MinGW compile for FFMPEG-based projects (Windows)
+* Change: New installation properties default video/dvd_video renderer is 'EVR'
+* Change: New installation properties default video/audio decoder is 'Default'
```
Narflex:
```suggestion
* Change: New installation properties default video/audio decoder are auto-detected
```

Narflex:
Just one change to the release notes...otherwise I'm fine with it.

CHANGELOG.md
wnjj: I'll add that and the Win10 VISTA_OS detection to the change log.

PR #409, opened 2019-03-07
Fix 64-bit DVD menu.

wnjj:
Due to an improper data type for the params passed into GetEvent(), DVD menus wouldn't work on Windows 64-bit version.  I added (int, UNLONG and DWORD) casts to suppress compiler warnings.  The fix has been confirmed by me and another user.


Narflex:
Thanks!

PR #410, opened 2019-03-12
Fix Windows 64-bit music visualizer and AVI playback crashes.

wnjj:
This fixes 2 more 64-bit Windows issues:  1. The music visualizer would crash when started.  2. AVI playback would crash on some systems.  The music fix needed a 64-bit specific DLL function call plus a struct 'pack' pragma to align with the Java DLL's expectation.  I added the same 'pack' to the other place we use the same struct even though it was already defaulting to the right size, in case a future code change affects that default.  The AVI fix was in the File Source filter where downstream filter context reference needed to be a DWORD_PTR instead of a DWORD.  Tested by me and 2 other users.

### DIFF
--- a/native/dll/DShowPlayer-2/MusicVisualizer.cpp
+++ b/native/dll/DShowPlayer-2/MusicVisualizer.cpp
@@ -22,7 +22,8 @@
 #include "sage_DShowMediaPlayer.h"
 #include "imusicvis.h"
 #include "guids.h"
-#include <jawt_md.h>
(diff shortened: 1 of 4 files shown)


wnjj:
Issue discussion here:  https://forums.sagetv.com/forums/showthread.php?p=620221&postcount=467

Narflex:
Travis said the build passed..not sure why it's still telling me in gitHub it's pending. I'll merge it. Thanks for the fix!

PR #416, opened 2019-09-06
Fix 64-bit Windows service launcher.  Set default heap to 1G for 64-bit.

wnjj:
This fixes a 64-bit automatic service startup issue discussed here:  https://forums.sagetv.com/forums/showthread.php?t=66069

Also increased the heap to 1G for 64-bit installs since that's a major reason for switching to the 64-bit version.

### DIFF
--- a/native/exe/SageLauncher/launcher.cpp
+++ b/native/exe/SageLauncher/launcher.cpp
@@ -22,6 +22,12 @@
 #include <direct.h>
 #include <windows.h>
 
+#ifdef _WIN64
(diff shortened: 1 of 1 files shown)


wnjj:
Travis is broken and apparently others have run into the same issue.  This is beyond my skill set but here’s some discussion:  https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/9

PR #419, opened 2019-11-12
Schedules Direct EPG grabber failed to finish updating some satellite-based lineups

enternoescape:
This fixes a rare issue that currently only plagues one user. Basically the integer from Schedules Direct is an unsigned integer, but Java only does signed. The easiest way to work around this is to either not import the frequencyHz variable from their JSON reply or make it a long integer. I chose the later since the extra data might be useful down the road even though we don't use it now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Fix 64-bit service launcher (Windows)
 * Set default 1G heap for 64-bit (Windows)
 * Updates for OPTUS D1 transponder changes to DVB-S2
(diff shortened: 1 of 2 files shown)


Narflex:
I'm curious how the data gets into this object because I don't see where it's getting set. :)

enternoescape:
It's magic. :) GSON does some tricks to set the variables without the need for any exposed set methods. I can show you the exact path to how we get here if you really want.

Narflex:
I'm guessing it's using reflection and overriding the accessor....I've done
that before (just not in SageTV) 🙂

Jeffrey Kardatzke

PR #420, opened 2019-11-15
Fix Haupauge HVR55xx tuners with 3 BDA tuners

JREkiwi:


native/dll/DShowCapture-2/GraphManager.cpp
```diff
@@ -133,15 +137,26 @@ JNIEXPORT jlong JNICALL Java_sage_DShowCaptureDevice_initGraph0
 		rvs[1] = CreateDshowCaptureInfo( env, videoCaptureFilterName, capDevNum, deviceCaps, tvType, 
 			                                 TuningMode, Country );
 
+		strncpy(TuningMode, GetBDAType(deviceCaps, 2), sizeof(TuningMode));
+		slog((env, "Create first DshowCapture %s.\r\n", TuningMode));
+		rvs[0] = CreateDshowCaptureInfo(env, videoCaptureFilterName, capDevNum, deviceCaps, tvType,
```
Narflex: Shouldn't this be rvs[2]? And then for the next block, rvs[3]? (comments also need updating from the copy/paste) this looks like it'd probably be better in a for loop from 0-3.

native/dll/DShowCapture-2/GraphManager.cpp
```diff
@@ -133,15 +137,26 @@ JNIEXPORT jlong JNICALL Java_sage_DShowCaptureDevice_initGraph0
...
+			TuningMode, Country);
+
+
+		strncpy(TuningMode, GetBDAType(deviceCaps, 3), sizeof(TuningMode));
+		slog((env, "Create second DshowCapture %s.\r\n", TuningMode));
+		rvs[1] = CreateDshowCaptureInfo(env, videoCaptureFilterName, capDevNum, deviceCaps, tvType,
+			TuningMode, Country);
+
```
Narflex: If you understand what the below block is doing can you add a comment to that effect? I'm not really following what this did before...or what the changes are modifying. This was some confusing code. :)

native/dll/DShowCapture-2/GraphManager.cpp
JREkiwi: Yes,  missed that.  Actually I see a few other things I need to tidy up in there

JREkiwi: I'll need to do some reading on how to do this as a loop

native/dll/DShowCapture-2/GraphManager.cpp
JREkiwi: Don't understand how it works but I can see there's a couple of copy/paste errors and a bit missing in the commit.  Looks like it's creating the ?table? to be used for AddBDAVideoCaptureFilters later in the code
Also a bit disconcerting all the //ZQ REMOVE ME comments through the code

wnjj:
From my brief look at that section of code it seems to be building a data structure with the first 2 BDA tuner types supported by the device, in priority order of ATSC, DVB_T, DVB_C, DVB_S, QAM.  Then it returns the first one as the default but there are pointers to the other one(s) contained within the data structure.

So is the idea with this commit to support more than 2 formats?  The comment edits suggest it is to support all 3 DVB formats so I suspect so.

I agree with Narflex's comment regarding loops.  There is a memory leak.  The currently proposed code only tears down graphs and frees memory for captures[0] and [1].  Those sections need updated to include the extra ones.  The code near the top of "Java_sage_DShowCaptureDevice_teardownGraph0" should tear down all of the referenced captures that aren't the current one before continuing on to tear down the current one explicitly.  Use a loop with the "dwBDAType != pCapInfo->dwBDAType" conditional for all 4.  Near the bottom of that same function a loop can be used to "delete" all 4 when done.

JREkiwi:
I'll need some more guidance on setting up as a loop. Copy/Paste and modify is pretty much the extent of my coding skill :)

native/dll/DShowCapture-2/GraphManager.cpp
```diff
@@ -362,8 +401,13 @@ JNIEXPORT void JNICALL Java_sage_DShowCaptureDevice_teardownGraph0
...
 		{
 			TeardownBDAGraph( env, pCapInfo->captures[1] );
+		} else
+		{
+			TeardownBDAGraph( env, pCapInfo->captures[2] );
+		} else
+		{
+			TeardownBDAGraph( env, pCapInfo->captures[3] );
 		}
-
```
JREkiwi: I need some help with this. I'm assuming this is where I need to do a loop as else are obviously not right here

JREkiwi: Does it need to something like this? Not sure if I should be testing against just num 4 or something else
```
		if ( pCapInfo->captures[0]->dwBDAType != pCapInfo->dwBDAType )
		for (int i = 0; i<4; i++)
		{
			TeardownBDAGraph( env, pCapInfo->captures[i] );
		} 

```

wnjj: For the teardown, put the compare inside the loop like this:

` 
		
	int i;

	for (i = 0; i<4; i++)
	{
		if ( pCapInfo->captures[i]->dwBDAType != pCapInfo->dwBDAType )
			TeardownBDAGraph( env, pCapInfo->captures[i] );
	}
`

Also needed [i] instead of [0] in the 'if'.  This will tear down all except the current one.

.
.
.

To loop the TuningMode assignment stuff consider this:

`

		DeviceCaps = deviceCaps & ~BDA_CAPTURE_ALL;
		
		int i,j;

		for (i=0; i<4; i++)
		{
			strncpy( TuningMode, GetBDAType( deviceCaps, i ), sizeof(TuningMode) );
			slog((env, "Create first DshowCapture %s.\r\n", TuningMode ));
			rvs[i] = CreateDshowCaptureInfo( env, videoCaptureFilterName, capDevNum, deviceCaps, tvType, 
				                                 TuningMode, Country );
		}


		for (i=0; i<4; i++)
		{
			if ( rvs[i] )
			{
				rvs[i]->captureNum = 4;
				for (j=0; j<4; j++)
					rvs[i]->captures[j] = rvs[j];
			}
		}

		rv = new DShowCaptureInfo;
		memset( rv, 0, sizeof(rv) );

		for (i=0; i<4; i++)
		{
			if ( rvs[i] )
			{
				memcpy( rv, rvs[i], sizeof(DShowCaptureInfo) );
				break;
			}
		}

		if (i==4)		// no valid rvs
			return 0;

		setChannelDev( (CHANNEL_DATA*)rv->channel, (void*)rv );
		return (jlong)rv;

`
.
.
.
Also don't forget to fix the 'deletes' at the bottom of 'Java_sage_DShowCaptureDevice_teardownGraph0':

`

	int i;
	for (i=0; i<4; i++)
	{
		if ( pCapInfo->captures[i] )
			delete pCapInfo->captures[i];
	}
`
.
.
.
Lastly, did you remember to change this line in DShowCapture.h?

`#define CAPTURE_NUM 4`
That sets the size for the 'captures' data structures.

JREkiwi: Thanks wnjj. I really appreciate you help. It does emphasize to me how little I know about this though.:)
I've made the changes but now SageTV crashes on startup when it gets to the tuningmode assignment code, so just trying to debug that.

JREkiwi:
Looks like I need to rethink/rework this, as the fixes for my initial changes for the number of tuners just cause SageTV to crash on initialization.  The 3 BDA tuners on the HVR-5525 are actually split over 2 capture devices, but currently the only way I can get it to work is to add all 3 to both capture devices. If I can identify how to code it for DVB-S on the Alt TS Capture and DVB-T, DVB-C on TS Capture then I wouldn't need to change the max tuners

JREkiwi:
I may as well close this and do another pull when I have these issues sorted

PR #423, opened 2019-12-06
Fix for Hauppauge 885 tuners with Alt TS Capture Devices (Windows)

JREkiwi:
Each TS capture device now correctly only displays the tuners available on that capture device.   Analog capture still needs work

### DIFF
--- a/native/dll/DShowCapture-2/DeviceVerify.cpp
+++ b/native/dll/DShowCapture-2/DeviceVerify.cpp
@@ -842,7 +842,7 @@ JNIEXPORT jint JNICALL Java_sage_DShowCaptureDevice_getDeviceCaps0
 	//Processing Hauppauge tuners
 	if ( hcw && bFoundDevice )
 	{
-		slog((env, "It's Hauppauge device.\r\n") );
+		slog((env, "It's a Hauppauge device.\r\n") );
 		if ( strstr( CaptureDrvInfo.device_desc, "HVR-1250") || strstr( CaptureDrvInfo.device_desc, "HVR-1255") || 
 			 strstr( CaptureDrvInfo.device_desc, "HVR-1550") )
 		{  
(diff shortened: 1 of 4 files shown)


native/dll/DShowCapture-2/DeviceVerify.cpp
```diff
@@ -876,32 +876,84 @@ JNIEXPORT jint JNICALL Java_sage_DShowCaptureDevice_getDeviceCaps0
...
+		// This still needs more work as the Analog capture doesn't work
+		if (strstr(CaptureDrvInfo.device_desc, "Model 121xxx, Hybrid DVB-T/S2") || strstr(CaptureDrvInfo.device_desc, "Model 53xxx, Hybrid DVB-T/S"))
 		{
 			isHybridCapture = false;
 			hasCaptureDetails = false;
 			detailsCaptureMask = 0x000800;  // sage_DShowCaptureDevice_MPEG_AV_CAPTURE_MASK
 			detailsChipsetMask = 0;
-			BDAInputType = sage_DShowCaptureDevice_BDA_DVB_S | sage_DShowCaptureDevice_BDA_DVB_T;
-			slog((env, "A bundle card is found (Analog+DVB-T+DVB-S), DVB-T|DVB-S added into source.\r\n"));
+			if (strstr(capFiltName, "Hauppauge WinTV 885 Alt TS Capture"))
```
Narflex: This code block is duplicated below, so it should be abstracted out into a function to avoid code duplication. You should also handle the case where neither capFiltName matches and then set some kind of default BDAInputType...or at the least print out some kind of error so we know that case was hit.

JREkiwi: Done. Not sure if this is exactly how you wanted it.
Didn't really want to get too generic as I didn't want to break stuff that is working

Narflex: Sorry...I didn't realize there was a subtle difference in the handling in each path...so forget what I recommended there...EXCEPT for the part about logging something if we hit that last else case.

JREkiwi: I'm sure my comments and logging info didn't help so I've modified that and removed the functions.
We always do hit the last else case, but we don't need to do anything with anything other than the specified capFiltNames, but have added logging for when we hit it.

Narflex:
Looks fine, thanks!

PR #425, opened 2020-01-02
Clean up of VS2015 warnings - Part 1

JREkiwi:
Fix C6283: <variable> is allocated with array new [], but deleted with scalar delete
Fix C4244: conversion from 'int' to 'size_t' ', possible loss of data for length variables
Remove redundant  #include <olectlid.h>


JREkiwi:
native/exe/HCWIRBlasterCOM/HCWIRBlasterCOM_p.c  shouldn't have been included. Can I just delete that from here?

wnjj:
> native/exe/HCWIRBlasterCOM/HCWIRBlasterCOM_p.c shouldn't have been included. Can I just delete that from here?

Yes, you can delete it and then commit again.

wnjj:
Be careful when changing data types just to make warnings go away.  I reviewed most all of them and decided at the time to leave them.  Most are for things like string lengths that will never overflow an int even though 64-bit technically allows for strings larger than that.  The concern I had (and why I left them alone) is that many of those lengths get passed into lower level functions expecting “int”.  Passing size_t into those may mess up the stack unless the compiler corrects for it somehow (or you add casts to int).  Either way it’s being forced back into an int anyway which means the max length is still the same.

Another case I saw where “n = ...” got changed to “size_t n = ...”. This is actually creating a new local scoped variable that wasn’t local before and creates a new risk if there’s still an “int n” somewhere in the neighborhood which will now be a distinct variable.

I’m not saying it can’t be done but that each and every one should be closely examined.

native/ax/StrmDeMux/SDeMuxFilter.cpp
```diff
@@ -2414,7 +2414,8 @@ void CSDeMuxFilter::dump_data( char* pData, int dwBytes )
 void CSDeMuxFilter::dump_sample( IMediaSample *pSample )
 {
 	BYTE* pBuffer;
-	int   Length;
+	size_t Length = 0;
+	size_t offset = 0;
```
Narflex: Change 'offset' in the header...or just get rid of it altogether since it doesn't look like it's used for anything.

native/ax/StrmDeMux/SDeMuxInPin.cpp
```diff
@@ -513,7 +513,8 @@ void CSDeMuxInPin::dump_data( char* pData, int dwBytes )
 void CSDeMuxInPin::dump_sample( IMediaSample *pSample )
 {
 	BYTE* pBuffer;
-	int   Length;
+	size_t Length = 0;
+	size_t offset = 0;
```
Narflex: Same thing for offset here

native/dll/DShowCapture-2/ScanChannel.cpp
```diff
@@ -139,7 +139,7 @@ JNIEXPORT jstring JNICALL Java_sage_DShowCaptureDevice_scanBDAChannel0
 			   
 	char* scanResult = scanChannel( (CHANNEL_DATA*)pCapInfo->channel, deviceName, tuningString,  
 		                               countryRegion,  (int)streamType );
-	len = strlen( scanResult );
+	(size_t) len = strlen( scanResult );
```
Narflex: Not sure why you added the cast on the left hand side here...len should already be a size_t.

native/dll/WinRawInput/WinRawInput.cpp
```diff
@@ -159,7 +159,7 @@ static WORD GetRawInputCode( void* RawInData, BYTE* pCode )
 			ret = raw->data.hid.bRawData[1];
 	}
 
-	delete lpb;
+	delete [] lpb;
```
Narflex: Shouldn't you also change line 150?

native/exe/HCWIRBlasterCOM/HCWIRBlasterCOM_p.c
```diff
@@ -0,0 +1,354 @@
+
```
Narflex: Remove this file as you mentioned....

native/exe/MiniClientLauncher/MiniClientLauncher.cpp
```diff
@@ -149,8 +149,8 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	HMODULE exeMod = GetModuleHandle(NULL);
 	LPTSTR appPath = new TCHAR[512];
 	GetModuleFileName(exeMod, appPath, 512);
-	int appLen = strlen(appPath);
-	for (int i = appLen - 1; i > 0; i--)
+	size_t appLen = strlen(appPath);
+	for (size_t i = appLen - 1; i > 0; i--)
```
Narflex: This is dangerous if appLen is zero...so handle that case too.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -625,8 +625,8 @@ int WINAPI WinMain( HINSTANCE hInst, 	/*Win32 entry-point routine */
 	// Fix the current working directory to be where the EXE is located
 	LPTSTR folderPath = new TCHAR[2048];
 	GetModuleFileName(NULL, folderPath, 2048);
-	int appLen = strlen(folderPath);
-	for (int i = appLen - 1; i > 0; i--)
+	size_t appLen = strlen(folderPath);
+	for (size_t i = appLen - 1; i > 0; i--)
```
Narflex: Again, dangerous if appLen is zero (which I know shouldn't happen, but still...it's good coding).

native/exe/SageLauncher/launcher.cpp
```diff
@@ -919,8 +919,8 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
 	
 	LPTSTR appPath = new TCHAR[2048];
 	GetModuleFileName(NULL, appPath, 2048);
-	int appLen = strlen(appPath);
-	for (int i = appLen - 1; i > 0; i--)
+	size_t appLen = strlen(appPath);
+	for (size_t i = appLen - 1; i > 0; i--)
```
Narflex: Same thing about appLen being zero.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -180,8 +180,8 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	HMODULE exeMod = GetModuleHandle(NULL);
 	LPTSTR appPath = new TCHAR[512];
 	GetModuleFileName(exeMod, appPath, 512);
-	int appLen = strlen(appPath);
-	for (int i = appLen - 1; i > 0; i--)
+	size_t appLen = strlen(appPath);
+	for (size_t i = appLen - 1; i > 0; i--)
```
Narflex: Ditto

third_party/Elecard/MpegAudioEncLayer2/Mpegac.cpp
```diff
@@ -22,7 +22,6 @@
 #include <streams.h>
 #include <olectl.h>
 #include <initguid.h>
-#include <olectlid.h>
```
Narflex: Why are you removing this include?  I'd prefer to leave this third party code alone.

third_party/Elecard/MpegAudioEncLayer2/utils.cpp
```diff
@@ -16,6 +16,7 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
+#define _CRT_SECURE_CPP_OVERLOAD_SECURE_NAMES 1
```
Narflex: I'd prefer to leave this one alone...

JREkiwi:
Thanks, from my reading it appeared that size_t should be used for string lengths as strnlen() always returns a size_t as it should never be a negative.
I'll recheck for the “size_t n = ...” as I may have been over zealous.

third_party/Microsoft/Dump/MPEG2Dump.cpp
```diff
@@ -1282,7 +1282,7 @@ HRESULT CMPEG2Dump::WriteNetwork(PBYTE data, DWORD len)
 		temp[--pos] = (TCHAR) digit+L'0';
 	} while (li.QuadPart);
 	sprintf(request, "WRITE %s %d\r\n", temp+pos, m_bRegOptimizeTransfers ? m_bytesLeftInNetworkWrite : len);
-	int dataSize = strlen(request);
+	size_t dataSize = strlen(request);
```
Narflex: Did this generate a compiler warning for the next line regarding a signed vs. unsigned comparison?

wnjj:
> Thanks, from my reading it appeared that size_t should be used for string lengths as strnlen() always returns a size_t as it should never be a negative.
> I'll recheck for the “size_t n = ...” as I may have been over zealous.

Sorry my reply was from a mobile device and I didn't go back to find the location.  It's at line 3509 of native/ax/Channel-2/Channel.c which looks like it's only used very locally.  You should probably remove 'n' from the int declaration on line 3496 if you haven't already.

I agree that size_t is the proper type for strlen because it's unsigned but also because it grows to 64-bits for 64-bit compiles.  My concern is where that now 64-bit value gets sent back into a 32-bit int location.  I would hope the compiler would report warnings where there is danger so as long as they are all reviewed or better yet all gone it should be ok.

Also, is there anywhere in the code where two strlen's are subtracted which may result in a negative number (as int's) that now may behave differently when unsigned?

native/ax/StrmDeMux/SDeMuxFilter.cpp
JREkiwi: Line 2424 complains of size_t to int conversion for offset

Changed in header

native/ax/StrmDeMux/SDeMuxInPin.cpp
JREkiwi: Done.

native/dll/DShowCapture-2/ScanChannel.cpp
JREkiwi: Removed, got added before I did the change at Line 84

native/dll/WinRawInput/WinRawInput.cpp
JREkiwi: Yes, had already caught that

native/exe/HCWIRBlasterCOM/HCWIRBlasterCOM_p.c
JREkiwi: Done

third_party/Elecard/MpegAudioEncLayer2/Mpegac.cpp
JREkiwi: olectlid.h just contains

#pragma message("WARNING: your code should #include olectl.h instead of olectlid.h")
#include <olectl.h>

third_party/Microsoft/Dump/MPEG2Dump.cpp
JREkiwi: On review send requires an int and size of request is 512
Reverted to int also @ 1680 and 1704
Added comments

wnjj:
You can probably just cast the strlen return value to int to suppress the type warning.
-------- Original message --------From: JREkiwi <notifications@github.com> Date: 1/2/20  12:42 PM  (GMT-08:00) To: google/sagetv <sagetv@noreply.github.com> Cc: Warren Juenemann <warren.juenemann@gmail.com>, Comment <comment@noreply.github.com> Subject: Re: [google/sagetv] Clean up of VS2015 warnings - Part 1 (#425) @JREkiwi commented on this pull request.



In third_party/Microsoft/Dump/MPEG2Dump.cpp:
> @@ -1282,7 +1282,7 @@ HRESULT CMPEG2Dump::WriteNetwork(PBYTE data, DWORD len)
 		temp[--pos] = (TCHAR) digit+L'0';
 	} while (li.QuadPart);
 	sprintf(request, "WRITE %s %d\r\n", temp+pos, m_bRegOptimizeTransfers ? m_bytesLeftInNetworkWrite : len);
-	int dataSize = strlen(request);
+	size_t dataSize = strlen(request);

On review send requires an int and size of request is 512
Reverted to int also @ 1680 and 1704
Added comments

—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/google/sagetv/pull/425?email_source=notifications\u0026email_token=AHYS4NE5CAS62ALFDZOKB3DQ3ZGU3A5CNFSM4KB5YKU2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCQRSU3A#discussion_r362622583",
"url": "https://github.com/google/sagetv/pull/425?email_source=notifications\u0026email_token=AHYS4NE5CAS62ALFDZOKB3DQ3ZGU3A5CNFSM4KB5YKU2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCQRSU3A#discussion_r362622583",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

native/exe/MiniClientLauncher/MiniClientLauncher.cpp
JREkiwi: Think I have this fixed appropriately

native/exe/SageLauncher/launcher.cpp
JREkiwi: Ditto fix

native/exe/SageLauncher/launcher.cpp
JREkiwi: Ditto Fix

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Ditto fix

JREkiwi:
Was looking at casting the strlen return value to int.  I'm sure I'll have many more commits for this :)

Narflex:
Are you working on more changes to this or is it ready to be reviewed again?

JREkiwi:
Won't get any more changes done today, so OK for review now.  Thanks

native/exe/MiniClientLauncher/MiniClientLauncher.cpp
```diff
@@ -150,7 +150,7 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	LPTSTR appPath = new TCHAR[512];
 	GetModuleFileName(exeMod, appPath, 512);
 	size_t appLen = strlen(appPath);
-	for (size_t i = appLen - 1; i > 0; i--)
+	for (size_t i = appLen; i-- > 0;) // this does the check for continuation on the value before decrement, but still uses the decremented value inside the loop
```
Narflex: A much cleaner solution is to just put that for loop inside an 'if (appLen > 0)' conditional block...then you don't need an ugly comment like this. :)

native/exe/SageLauncher/launcher.cpp
```diff
@@ -626,7 +626,7 @@ int WINAPI WinMain( HINSTANCE hInst, 	/*Win32 entry-point routine */
 	LPTSTR folderPath = new TCHAR[2048];
 	GetModuleFileName(NULL, folderPath, 2048);
 	size_t appLen = strlen(folderPath);
-	for (size_t i = appLen - 1; i > 0; i--)
+	for (size_t i = appLen; i-- > 0;) // this does the check for continuation on the value before decrement, but still uses the decremented value inside the loop
```
Narflex: Ditto

native/exe/SageLauncher/launcher.cpp
```diff
@@ -920,7 +920,7 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
 	LPTSTR appPath = new TCHAR[2048];
 	GetModuleFileName(NULL, appPath, 2048);
 	size_t appLen = strlen(appPath);
-	for (size_t i = appLen - 1; i > 0; i--)
+	for (size_t i = appLen; i-- > 0;) // this does the check for continuation on the value before decrement, but still uses the decremented value inside the loop
```
Narflex: Ditto

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -181,7 +181,7 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	LPTSTR appPath = new TCHAR[512];
 	GetModuleFileName(exeMod, appPath, 512);
 	size_t appLen = strlen(appPath);
-	for (size_t i = appLen - 1; i > 0; i--)
+	for (size_t i = appLen; i-- > 0;) // this does the check for continuation on the value before decrement, but still uses the decremented value inside the loop
```
Narflex: Ditto

third_party/Microsoft/Dump/MPEG2Dump.cpp
```diff
@@ -1282,7 +1282,8 @@ HRESULT CMPEG2Dump::WriteNetwork(PBYTE data, DWORD len)
 		temp[--pos] = (TCHAR) digit+L'0';
 	} while (li.QuadPart);
 	sprintf(request, "WRITE %s %d\r\n", temp+pos, m_bRegOptimizeTransfers ? m_bytesLeftInNetworkWrite : len);
-	size_t dataSize = strlen(request);
+// this throws a warning C4267, but send needs int
+	int dataSize = strlen(request);
```
Narflex: You can just put an (int) cast in there to suppress the warning.

third_party/Microsoft/Dump/MPEG2Dump.cpp
```diff
@@ -1675,7 +1676,8 @@ HRESULT CMPEG2Dump::OpenConnection()
 	char data[512];
 	sprintf(data, "WRITEOPEN %s %d\r\n", lpszFileName, m_dwUploadKey);
 	delete [] lpszFileName;
-	size_t dataSize = strlen(data);
+// this throws a warning C4267, but send needs int
+	int dataSize = strlen(data);
```
Narflex: Ditto

third_party/Microsoft/Dump/MPEG2Dump.cpp
```diff
@@ -1698,7 +1700,8 @@ HRESULT CMPEG2Dump::CloseConnection()
 {
 	DbgLog((LOG_TRACE, 2, TEXT("CloseConnection() IN")));
 	char* data = "QUIT\r\n";
-	size_t dataSize = strlen(data);
+// this throws a warning C4267, but send needs int
+	int dataSize = strlen(data);
```
Narflex: Ditto

third_party/Microsoft/Dump/MPEG2Dump.cpp
JREkiwi: Done

third_party/Microsoft/Dump/MPEG2Dump.cpp
JREkiwi: Done

third_party/Microsoft/Dump/MPEG2Dump.cpp
JREkiwi: Done

native/exe/MiniClientLauncher/MiniClientLauncher.cpp
JREkiwi: Was tossing up whether to just put it inside an 'if (appLen > 0)' conditional, then I read [this](https://stackoverflow.com/questions/1951519/should-i-use-stdsize-t-or-int-in-my-for-loops)  :)
Changed to  'if (appLen > 0)' conditional

native/exe/SageLauncher/launcher.cpp
JREkiwi: Changed to  'if (appLen > 0)' conditional

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Changed to  'if (appLen > 0)' conditional

JREkiwi:
Just as an aside, I've been running with the majority of these changes for around a week and was thinking I was going a bit crazy as my system seemed to be performing a bit snappier, which I was sure must be related to something else. Then this morning I stumbled across [this](https://hownot2code.com/2016/10/10/part-25-optimization-of-64-bit-programs/) and some very interesting performance comparisons.

wnjj:
I checked that link on code optimization and I’m sure there’s something to it.  That said, most of SageTV’s performance snappiness is dependent upon the Java code.  Furthermore much of the changes are to string processing of file names, etc. and aren’t executing frequently enough to be noticed so I don’t see how you would notice but I suppose it’s possible.

Regardless, thank you for taking this on as clean code is easier to support in the future and having tons of “known” warnings make catching new mistakes more difficult.

Also, all of those (size_t) casts from strlen() in native/ax/TSnative/SIParser.c can be removed If you’re going to let it do the math as size_t but doesn’t the compiler complain that it’s now comparing the results to an int (the variable called “Size”)?  Does the “+ 1” cause the expression to cast to int anyway?

JREkiwi:
I couldn't see why the changes would make any difference to performance and I was pretty sure it must be something else, but it made for interesting reading as I was researching how to do some of these changes.
I do need to fix up native/ax/TSnative/SIParser.c as the changes aren't doing anything useful

JREkiwi:
Reverted the changes to native/ax/TSnative/SIParser.c as I need to do a bit more learning before I tackle it again.

native/exe/SageLauncher/launcher.cpp
JREkiwi: Changed to 'if (appLen > 0)' conditional

native/exe/MiniClientLauncher/MiniClientLauncher.cpp
Narflex: Looks like you missed this one...

JREkiwi: And that's why there needs to be at least another pair of eyes looking at it :)
Done. Thanks

JREkiwi:
Unless there's something else I've missed, there's nothing else for this pull
Haven't decided what to focus on for Part 2, probably conversion from '__int64' to 'int'
Thanks

Narflex:
Thanks!

PR #429, opened 2020-01-26
Revert changes to CPushReader.cpp

JREkiwi:
:)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1017,11 +1017,3 @@ native/.vs/SageWorkspace/v14/.suo
 *.log
 native/SageWorkspace.VC.db
 native/SageWorkspace.VC.VC.opendb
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1017,3 +1017,11 @@ native/.vs/SageWorkspace/v14/.suo
 *.log
 native/SageWorkspace.VC.db
 native/SageWorkspace.VC.VC.opendb
+third_party/Elecard/MpegAudioEncLayer2/Win32/Release/allcodecs.obj
```
Narflex: If you want to change .gitignore, please do that in a separate change. And if you do, don't put .lib files in that list because some of those are actually in GitHub for reasons I can't remember. :)

JREkiwi: removed .gitignore

Narflex:
Thanks!

PR #430, opened 2020-02-03
Change service launcher (Windows) to support local JRE

JREkiwi:
Support local SageTV folder JRE
Tidy up warnings in VS2015 by changing to MS "safe" APIs
Add ReleaseStringUTFChars to release allocated memory
Change JAVA version in popup messages
Remove disable warnings pragma

### DIFF
--- a/native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
+++ b/native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
@@ -264,8 +264,8 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 //env->ExceptionDescribe();
 		jstring throwStr = (jstring)env->CallObjectMethod(mainThrow, toStr);
 		const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
-		char *errStr = new char[env->GetStringLength(throwStr) + 64];
-		size_t errStrlen = strlen(cThrowStr) + 64;  // Need to know this size for the sprintf_s
+		size_t errStrlen = env->GetStringLength(throwStr) + 64;
+		char *errStr = (char*)malloc(errStrlen);
 		sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
 		errorMsg(errStr, "Java Exception");
 		delete [] errStr;
--- a/native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
+++ b/native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
@@ -103,31 +103,23 @@ void popupExceptionError(JNIEnv* env, jthrowable thrower, char* errTitle)
 	const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
 	size_t errStrlen = env->GetStringLength(throwStr) + 64;
 	char *errStr = (char*) malloc(errStrlen);
-	sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
-	errorMsg(errStr, errTitle);
-	free(errStr);
+	if (errStr != nullptr)  // Should never be NULL
+	{
+		sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
+		errorMsg(errStr, errTitle);
+		free(errStr);
+	}
 	env->ReleaseStringUTFChars(throwStr, cThrowStr);
 }
 
 
-int APIENTRY WinMain(HINSTANCE hInstance,
-                     HINSTANCE hPrevInstance,
-                     LPSTR     lpCmdLine,
-                     int       nCmdShow)
+int APIENTRY WinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance, _In_ LPSTR lpCmdLine, _In_ int nCmdShow)
 {
 	/*
 	 * Explicitly load jvm.dll.
 	 */
-	HKEY rootKey = HKEY_LOCAL_MACHINE;
-	char currVer[16];
-	HKEY myKey;
-	DWORD readType;
-	DWORD dwRead = 0;
-	DWORD hsize = sizeof(dwRead);
-	hsize = sizeof(currVer);
-	HMODULE exeMod = GetModuleHandle(NULL);
-	char appPath[512];
-	GetModuleFileName(exeMod, appPath, sizeof(appPath));
+	char appPath[_MAX_PATH];
+	GetModuleFileName(NULL, appPath, sizeof(appPath));
 	size_t appLen = strlen(appPath);
 	if (appLen > 0)  // Shouldn't be 0 as the following would be an infinite loop
 	{
@@ -142,7 +134,7 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	}
 
 	// See if we've got a JVM in our own directory to load
-	char includedJRE[1024];
+	char includedJRE[_MAX_PATH];
 	strcpy_s(includedJRE, sizeof(includedJRE), appPath);
 	strcat_s(includedJRE, sizeof(includedJRE), "jre\\bin\\client\\jvm.dll");
 	HMODULE hLib = LoadLibrary(includedJRE);
@@ -150,6 +142,11 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	if (hLib == NULL)
 	{
 		// Failed to find JRE in SageTV directory, load the JVM from the registry instead, by using the Windows Registry to locate the current version to use.
+		HKEY rootKey = HKEY_LOCAL_MACHINE;
+		char currVer[16];
+		HKEY myKey;  // myKey to NULL for RegQueryValueEx
+		DWORD readType;  // readType to NULL for RegQueryValueEx
+		DWORD hsize = sizeof(currVer);
 
 		if (RegOpenKeyEx(rootKey, "Software\\JavaSoft\\Java Runtime Environment", 0, KEY_QUERY_VALUE, &myKey) != ERROR_SUCCESS)
 		{
@@ -167,12 +164,12 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 		char pathKey[1024];
 		strcpy_s(pathKey, sizeof(pathKey), "Software\\JavaSoft\\Java Runtime Environment\\");
 		strcat_s(pathKey, sizeof(pathKey), currVer);
-		char jvmPath[1024];
 		if (RegOpenKeyEx(rootKey, pathKey, 0, KEY_QUERY_VALUE, &myKey) != ERROR_SUCCESS)
 		{
 			JVM_MISSING;
 			return FALSE;
 		}
+		char jvmPath[_MAX_PATH];
 		hsize = sizeof(jvmPath);
 		if (RegQueryValueEx(myKey, "RuntimeLib", 0, &readType, (LPBYTE)jvmPath, &hsize) != ERROR_SUCCESS)
 		{
@@ -204,11 +201,16 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	}
 
 	// Set the current working directory to be the folder the EXE is in.
-	_chdir(appPath);
+	errno_t err;
+	if ((err = _chdir(appPath)) != 0) // If this returns 0 we have bigger problems
+	{
+		errorMsg("Could set the working directory", "Change to working directory failed");
+		return FALSE;
+	}
 
 	// Set up the JAVA VM
-	char jarPath[1024];
-	char libraryPath[512];
+	char jarPath[_MAX_PATH];
+	char libraryPath[_MAX_PATH];
 	
     JNIEnv *env;       /* pointer to native method interface */
 	JavaVMInitArgs vm_args;
@@ -266,9 +268,12 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 		const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
 		size_t errStrlen = env->GetStringLength(throwStr) + 64;
 		char *errStr = (char*)malloc(errStrlen);
-		sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
-		errorMsg(errStr, "Java Exception");
-		delete [] errStr;
+		if (errStr != nullptr)  // Should never be NULL
+		{
+			sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
+			errorMsg(errStr, "Java Exception");
+			free(errStr);
+		}
 		env->ReleaseStringUTFChars(throwStr, cThrowStr);
 		return FALSE;
 	}
--- a/native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
+++ b/native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
@@ -16,11 +16,13 @@
 // ServiceControlLaunch.cpp : Defines the entry point for the application.
 
 /*
-*		*********************************************************
-*		Microsoft Visual Studio project codepage set to MBCS 
-*		as JAVA does not support being passed Unicode parameters
-*		*********************************************************
-*/
+ *       ******************************************************************************************
+ *       Microsoft Visual Studio project codepage intentionally set to MBCS
+ *       The strings passed in the JavaVMOption struct use the platform default character encoding,
+ *       so they can't be passed as 16-bit Unicode chars.
+ *
+ *       *******************************************************************************************
+ */
 
 #include "stdafx.h"
 
@@ -33,13 +35,6 @@
 
 #define JVM_MISSING MessageBox(NULL, "Could not get information on current JVM.\nPlease install Java Runtime Environment 1.7 or higher (Java 1.8 preferred)", "Java Missing", MB_OK);\
 
-#ifndef _MSC_VER // Cater for compiling not in MSVC
-#define strcpy_s(dst, dst_length, src) strcpy(dst, src)
-#define strcat_s(dst, dst_length, src) strcat(dst, src)
-#define vsprintf_s(dst, dst_length, format, args) vsprintf(dst, format, args)
-#define sprintf_s(dst, dst_length, format, args) sprintf(dst, format, args)
-#endif
-
 static JNIEnv* globalenv = 0;
 
 JavaVM *vm;    
@@ -47,18 +42,17 @@ JavaVM *vm;
 void sysOutPrint(const char* cstr, ...)
 {
 	JNIEnv* env;
-	jint threadState = vm->GetEnv((void**)&env, JNI_VERSION_1_6); // JNI in JRE 1.6 and greater
+	jint threadState = vm->GetEnv((void**)&env, JNI_VERSION_1_2); // JNI in JRE 1.2 and greater
 	if (threadState == JNI_EDETACHED)
 		vm->AttachCurrentThread((void**)&env, NULL);
 	jthrowable oldExcept = env->ExceptionOccurred();
 	if (oldExcept)
 		env->ExceptionClear();
-    va_list args;
-    va_start(args, cstr);
-    int const bufsize = 1024;
-    char buf[bufsize];
-    vsprintf_s(buf, bufsize, cstr, args);
-    va_end(args);
+	va_list args;
+	va_start(args, cstr);
+	char buf[1024];
+	vsprintf_s(buf, sizeof(buf), cstr, args);
+	va_end(args);
 	jstring jstr = env->NewStringUTF(buf);
 	static jclass cls = (jclass) env->NewGlobalRef(env->FindClass("java/lang/System"));
 	static jfieldID outField = env->GetStaticFieldID(cls, "out", "Ljava/io/PrintStream;");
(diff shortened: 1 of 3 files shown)


JREkiwi:
I did find my answer to the Unicode vs MBCS project question. JAVA doesn't support being passed Unicode parameters and it became very messy when the project was Unicode. Much tidier keeping it as MBCS

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -14,29 +14,31 @@
...
-//
 
-#include "stdafx.h"
-
-
-// ServiceControlLauncher.cpp : Defines the entry point for the console application.
-//
+/*
+*		*********************************************************
+*		Microsoft Visual Studio project codepage set to MBCS
```
Narflex: Huh?  How could Java not support being passed Unicode parameters....Java's char primitive is a 16-bit unicode character.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -14,29 +14,31 @@
...
 #include <stdlib.h>
 #include <jni.h>
 #include <stdio.h>
 #include <direct.h>
 
 
-#define JVM_MISSING MessageBox(NULL, "Could not get information on current JVM.\nPlease install Java Runtime Environment 1.4","Java Missing", MB_OK);\
+#define JVM_MISSING MessageBox(NULL, "Could not get information on current JVM.\nPlease install Java Runtime Environment 1.7 or higher (Java 1.8 preferred)", "Java Missing", MB_OK);\
 
+#ifndef _MSC_VER // Cater for compiling not in MSVC
```
Narflex: This is not good practice to redefine 'safe' string functions to be their non-safe equivalent. There's no reason I can think this file wouldn't be compiled with MSVC either since it's making a Windows EXE.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -45,16 +47,17 @@ JavaVM *vm;
 void sysOutPrint(const char* cstr, ...)
 {
 	JNIEnv* env;
-	jint threadState = vm->GetEnv((void**)&env, JNI_VERSION_1_2);
+	jint threadState = vm->GetEnv((void**)&env, JNI_VERSION_1_6); // JNI in JRE 1.6 and greater
```
Narflex: If you're changing this to 1.6, you better provide a good reason as I have no idea what kind of breakage this may introduce by doing something like that.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -45,16 +47,17 @@ JavaVM *vm;
...
 	if (threadState == JNI_EDETACHED)
 		vm->AttachCurrentThread((void**)&env, NULL);
 	jthrowable oldExcept = env->ExceptionOccurred();
 	if (oldExcept)
 		env->ExceptionClear();
     va_list args;
     va_start(args, cstr);
-    char buf[1024];
-    vsprintf(buf, cstr, args);
+    int const bufsize = 1024;
```
Narflex: This block should be:

char buf[1024];
vsnprintf(buf, sizeof(buf), cstr, args);

That should be portable (not that we need it here) and less lines of code and uses safe string printing.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -105,9 +109,11 @@ void popupExceptionError(JNIEnv* env, jthrowable thrower, char* errTitle)
 	jstring throwStr = (jstring)env->CallObjectMethod(thrower, toStr);
 	const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
 	char *errStr = (char*) malloc(env->GetStringLength(throwStr) + 64);
-	sprintf(errStr, "An exception occured in Java:\n%s", cThrowStr);
+	size_t errStrlen = strlen(cThrowStr) + 64;  // Need to know this size for the sprintf_s
```
Narflex: This code I wrote is so ugly...but I wrote it many years ago. :)  A cleaner way here is to establish the length before the call to malloc and then pass that length in there. Then you can reuse that length in snprintf to ensure buffer safety.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -105,9 +109,11 @@ void popupExceptionError(JNIEnv* env, jthrowable thrower, char* errTitle)
 	jstring throwStr = (jstring)env->CallObjectMethod(thrower, toStr);
 	const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
 	char *errStr = (char*) malloc(env->GetStringLength(throwStr) + 64);
-	sprintf(errStr, "An exception occured in Java:\n%s", cThrowStr);
+	size_t errStrlen = strlen(cThrowStr) + 64;  // Need to know this size for the sprintf_s
+	sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
 	errorMsg(errStr, errTitle);
 	free(errStr);
+	env->ReleaseStringUTFChars(throwStr, cThrowStr);
```
Narflex: Nice catch, this was a memory leak...although nothing that would ever accumulate. :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -126,44 +132,75 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	DWORD dwRead = 0;
 	DWORD hsize = sizeof(dwRead);
 	hsize = sizeof(currVer);
-	if (RegOpenKeyEx(rootKey, "Software\\JavaSoft\\Java Runtime Environment", 0, KEY_QUERY_VALUE, &myKey) != ERROR_SUCCESS)
+	HMODULE exeMod = GetModuleHandle(NULL);
```
Narflex: If you want a good learning project...what really SHOULD have been done here when I wrote it was to refactor this into a common source file that's used by all 3 launchers which can invoke the JVM (and other reused stuff like sysOutPrint). That avoids code duplication and will make them all consistent.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -173,60 +210,44 @@ int APIENTRY WinMain(HINSTANCE hInstance,
...
+	LPTSTR jarPath = new TCHAR[jarPathsize];
+	LPTSTR libraryPath = new TCHAR[libraryPathsize];
 	
     JNIEnv *env;       /* pointer to native method interface */
 	JavaVMInitArgs vm_args;
 	JavaVMOption options[32];
 	vm_args.nOptions = 0;
-	strcpy(jarPath, "-Djava.class.path=");
-	strcat(jarPath, appPath);
+	strcpy_s(jarPath, jarPathsize, "-Djava.class.path=");
```
Narflex: While using separate constants isn't totally bad...putting them inline with the allocations and then using sizeof is much cleaner.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Should probably read JavaVMOption doesn't support Unicode

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Yeah,  I think we'd have much bigger problems before this caused an issue :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
wnjj: From what I could find online, it's due to the way JNI_CreateJavaVM expects its arguments.  Someone here had the same issue https://github.com/LeeKamentsky/python-javabridge/issues/35.

I'll admit I don't have a good grasp of what all is involved but Oracle's documentation says "the option as a string in the default platform encoding".  (https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html).

Near as I can tell, it has to do with the way the JVM itself is compiled for Windows.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: I was running into issues with sizeof as jarPath isn't initialized until after strcpy_s

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
Narflex: Ahh OK, now I see.  So this comment is just explaining why it's already the way it is and not changing anything. That's fine then.  But please reword it to indicate that the strings passed in the JavaVMOption struct use the platform default character encoding, so they can't be passed as 16-bit Unicode chars.  There's no question you can do other string operations in Java on Unicode chars. :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
Narflex: The jarPath line and the libraryPath line should just be changed to:

char jarPath[1024];
char libraryPath[1024];

and then sizeof will work fine and you can remove the free statements at the end...man, this was such bad coding I did...ick. Or maybe there was some reason I was using dynamic instead of stack allocation...but you'll find out when you try to change it if you get some weird compiler errors. :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Removed

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: MS docs say vsnprintf() is not safe either, and we don't need to be portable :)
OK to do this instead
    char buf[1024];
    vsprintf_s(buf, sizeof(buf), cstr, args);

Narflex: I'd love to see a pointer to those docs, because that's just wrong unless MS implemented it in an unsafe way. :)  But what you're doing here is fine for windows code.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Not using anything in JNI beyond 1.2, so no requirement to change other than to be the latest version that is supported in Java 1.7.
No problem changing it back to 1.2

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: I did notice there's a lot of common code.  :)
I'll see if I can figure out how to do that for a future change

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-vsnprintf-vsnprintf-l-vsnwprintf-vsnwprintf-l?view=vs-2015

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -257,9 +265,11 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 		jstring throwStr = (jstring)env->CallObjectMethod(mainThrow, toStr);
 		const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
 		char *errStr = new char[env->GetStringLength(throwStr) + 64];
-		sprintf(errStr, "An exception occured in Java:\n%s", cThrowStr);
+		size_t errStrlen = strlen(cThrowStr) + 64;  // Need to know this size for the sprintf_s
+		sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
```
JREkiwi: Missed fixing this one

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -126,44 +125,71 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	DWORD dwRead = 0;
```
Narflex: Can you get rid of this line since dwRead is never actually used? And then get rid of that bad 'hsize' variable and inline the sizeof() calls whenever hsize is used...that'll be so much cleaner.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -126,44 +125,71 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	DWORD dwRead = 0;
 	DWORD hsize = sizeof(dwRead);
 	hsize = sizeof(currVer);
-	if (RegOpenKeyEx(rootKey, "Software\\JavaSoft\\Java Runtime Environment", 0, KEY_QUERY_VALUE, &myKey) != ERROR_SUCCESS)
+	HMODULE exeMod = GetModuleHandle(NULL);
```
Narflex: I'm pretty sure you can get rid of this line and then just pass NULL as the first param to GetModuleFileName.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -126,44 +125,71 @@ int APIENTRY WinMain(HINSTANCE hInstance,
 	DWORD dwRead = 0;
 	DWORD hsize = sizeof(dwRead);
 	hsize = sizeof(currVer);
-	if (RegOpenKeyEx(rootKey, "Software\\JavaSoft\\Java Runtime Environment", 0, KEY_QUERY_VALUE, &myKey) != ERROR_SUCCESS)
+	HMODULE exeMod = GetModuleHandle(NULL);
+	char appPath[512];
```
Narflex: Change the 512 to be _MAX_PATH to be technically correct. :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -126,44 +125,71 @@ int APIENTRY WinMain(HINSTANCE hInstance,
...
+				appPath[i + 1] = 0;
+				break;
+			}
+		}
 	}
-	
-	if (RegQueryValueEx(myKey, "CurrentVersion", 0, &readType, (LPBYTE)currVer, &hsize) != ERROR_SUCCESS)
+
+	// See if we've got a JVM in our own directory to load
+	char includedJRE[1024];
```
Narflex: Use _MAX_PATH again instead of 1024.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -126,44 +125,71 @@ int APIENTRY WinMain(HINSTANCE hInstance,
...
-		JVM_MISSING;
-		return FALSE;
-	}
-	hsize = sizeof(jvmPath);
-	if (RegQueryValueEx(myKey, "RuntimeLib", 0, &readType, (LPBYTE)jvmPath, &hsize) != ERROR_SUCCESS)
-	{
+		char pathKey[1024];
+		strcpy_s(pathKey, sizeof(pathKey), "Software\\JavaSoft\\Java Runtime Environment\\");
+		strcat_s(pathKey, sizeof(pathKey), currVer);
+		char jvmPath[1024];
```
Narflex: Use _MAX_PATH again here for jvmPath....registry values are different, so don't use _MAX_PATH for pathKey (leave it alone).

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -173,60 +199,42 @@ int APIENTRY WinMain(HINSTANCE hInstance,
...
-	}
-
 	// Set the current working directory to be the folder the EXE is in.
-	chdir(appPath);
+	_chdir(appPath);
 
-	LPTSTR jarPath = new TCHAR[1024];
-	LPTSTR libraryPath = new TCHAR[512];
+	// Set up the JAVA VM
+	char jarPath[1024];
```
Narflex: Use _MAX_PATH for both of these.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: The RegQueryValueEx where hsize is used all require LPDWORD.
Is it best to cast them or just tidy up the hsize definition?

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: Done

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
Narflex: MS is misleading in those docs...but whatever, it's fine with Windows specific calls here.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
Narflex: Whoops, my bad. I forgot that's why we had to use a temp variable like this. Do NOT just cast it as that'd be totally wrong and almost definitely crash things. It's expecting a pointer to the value because it wants to know the size on input, and then I *think* it also sets that value to the returned size on output (which is why it needs a pointer).  So just put it back the way it was.

JREkiwi: Didn't think it would be right trying to cast it, so I had just done a tidy up. :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
JREkiwi: They seem to use the terms "Safe" and "Secure" pretty interchangeably.

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
```diff
@@ -104,10 +101,12 @@ void popupExceptionError(JNIEnv* env, jthrowable thrower, char* errTitle)
 	jmethodID toStr = env->GetMethodID(env->GetObjectClass(thrower), "toString", "()Ljava/lang/String;");
 	jstring throwStr = (jstring)env->CallObjectMethod(thrower, toStr);
 	const char* cThrowStr = env->GetStringUTFChars(throwStr, 0);
-	char *errStr = (char*) malloc(env->GetStringLength(throwStr) + 64);
-	sprintf(errStr, "An exception occured in Java:\n%s", cThrowStr);
+	size_t errStrlen = env->GetStringLength(throwStr) + 64;
+	char *errStr = (char*) malloc(errStrlen);
+	sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
```
JREkiwi: Switched over to the Visual Studio 2015 (v140) Platform Toolset and ran code analysis and it complained about the sprintf_s with warning C6387: 'errStr' could be '0':  this does not adhere to the specification for the function 'sprintf_s'.
I can make the warning go away with
```
	if (errStr !=nullptr)
		sprintf_s(errStr, errStrlen, "An exception occured in Java:\n%s", cThrowStr);
```
but isn't errStr going to be NULL until the sprintf_s ?

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
wnjj: I don't see any purpose for dwRead.  I think those lines could just be changed to "DWORD hsize = sizeof(currVer);".  The LPDWORD type is fine since that's what "&hsize" is and as Jeff pointed out, that function wants a pointer to a DWORD.  It holds the size of the buffer (currVer) provided to the function and the function returns the actual byte count put into the buffer in hsize.  Refer to the lpcbData parameter here: https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa

JREkiwi: Thanks, that's what I've done, just didn't want to do another commit until I'd knew what I should be doing with sprintf_s @106
Trying to keep the commits down below 5 for this change :)

native/exe/ServiceControlLaunch/ServiceControlLaunch.cpp
wnjj: errStr is a pointer and is NULL until it points to a valid heap memory address (assigned via the 'new' operator).  The sprintf_s function does not allocate the memory.  The warning (and your valid fix) is to prevent passing that NULL pointer into sprintf_s in the (very rare) event that 'new' cannot allocate memory.

wnjj: Actually, you should wrap the sprintf_s, errMsg and delete lines all inside the nullptr check since none are valid with NULL.

JREkiwi: Done

Narflex:
Thanks!

JREkiwi:
I think this is it?

PR #431, opened 2020-02-15
Clean up of VS2015 warnings - Part 3

JREkiwi:
Changes from SageService Launcher applied to MiniClient Launcher
Resolve Compile and Static Analysis warnings
Change to MS safe/secure functions

### DIFF
--- a/native/common/jni-util.cpp
+++ b/native/common/jni-util.cpp
@@ -690,7 +690,11 @@ jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, cha
 	}
 	const char* cstr = env->GetStringUTFChars(jstr, NULL);
 	*rvCount = (unsigned long) MIN(rvlen, strlen(cstr));
+#ifdef WIN32
 	strncpy_s(rv, sizeof(rv), cstr, MIN(rvlen, strlen(cstr)));
+#else
+	strncpy(rv, cstr, MIN(rvlen, strlen(cstr)));
+#endif
 	env->ReleaseStringUTFChars(jstr, cstr);
 	return JNI_TRUE;
 }
--- a/native/common/jni-util.cpp
+++ b/native/common/jni-util.cpp
@@ -665,7 +665,7 @@ jboolean GetMapBoolValue(JNIEnv* env, jobject map, const char* optionName, jbool
 	}
 	return JNI_TRUE;
 }
-jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, char* rv, unsigned long rvlen, unsigned long* rvCount)
+jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, char* rv, unsigned long rvlen)
 {
 	jstring keyString = env->NewStringUTF(optionName);
 	static jclass mapClass = (jclass) env->NewGlobalRef(env->FindClass("java/util/Map"));
@@ -689,12 +689,8 @@ jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, cha
 		return JNI_FALSE;
 	}
 	const char* cstr = env->GetStringUTFChars(jstr, NULL);
-	*rvCount = (unsigned long) MIN(rvlen, strlen(cstr));
-#ifdef WIN32
-	strncpy_s(rv, sizeof(rv), cstr, MIN(rvlen, strlen(cstr)));
-#else
-	strncpy(rv, cstr, MIN(rvlen, strlen(cstr)));
-#endif
+	if (strlen(cstr) + 1 > rvlen) return JNI_FALSE;
+	if (snprintf(rv, rvlen, "%s", cstr) < 0) return JNI_FALSE ;
 	env->ReleaseStringUTFChars(jstr, cstr);
 	return JNI_TRUE;
 }
--- a/native/common/jni-util.cpp
+++ b/native/common/jni-util.cpp
@@ -13,8 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Eliminate silly MS compiler security warnings about using POSIX functions
-#pragma warning(disable : 4996)
 
 #include "../include/jni-util.h"
 #include <string.h>
@@ -692,7 +690,7 @@ jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, cha
 	}
 	const char* cstr = env->GetStringUTFChars(jstr, NULL);
 	*rvCount = (unsigned long) MIN(rvlen, strlen(cstr));
-	strncpy(rv, cstr, MIN(rvlen, strlen(cstr)));
+	strncpy_s(rv, sizeof(rv), cstr, MIN(rvlen, strlen(cstr)));
 	env->ReleaseStringUTFChars(jstr, cstr);
 	return JNI_TRUE;
 }
--- a/native/dll/DX93D/SageTVDX93D.cpp
+++ b/native/dll/DX93D/SageTVDX93D.cpp
@@ -317,8 +317,16 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 	slog((env, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone));
 	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ|FILE_MAP_WRITE, 0, 0, 0);
 	unsigned int* myData = (unsigned int*) myPtr;
-	if (myData == NULL)
+	if (myPtr == NULL || evtReady == NULL || evtDone == NULL)
+	{
+		if (myPtr != NULL)
+			UnmapViewOfFile(myPtr);
+		if (evtReady != NULL)
+			CloseHandle(evtReady);
+		if (evtDone != NULL)
+			CloseHandle(evtDone);
 		return;
+	}
 	slog((env, "Starting to read...0x%p\r\n", myPtr));
 	int configured = 0;
 	int j = 0;
@@ -328,14 +336,13 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 		j++;
 		// Decrease this wait time so that if there's case where MPlayer has to be forcibly quit we
 		// can exit this loop in a short amount of time
-		if (evtReady != NULL)
-			if (WAIT_OBJECT_0 != WaitForSingleObject(evtReady, 50))
-			{
-				if (env->GetLongField(jo, fid_pD3DObject) == 0)
-					break; // 3D system is dead
-				//printf("WAIT TIMEOUT EXPIRED!!!\r\n");
-				continue;
-			}
+		if (WAIT_OBJECT_0 != WaitForSingleObject(evtReady, 50))
+		{
+			if (env->GetLongField(jo, fid_pD3DObject) == 0)
+				break; // 3D system is dead
+			//printf("WAIT TIMEOUT EXPIRED!!!\r\n");
+			continue;
+		}
 		int currCmd = (myData[0] >> 24) & 0xFF;
 		//slog((env, "Got video cmd 0x%x\r\n", currCmd));
 		if (currCmd == 0x80)
@@ -367,8 +374,7 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 		else
 		{
 		}
-		if (evtDone != NULL)
-			SetEvent(evtDone);
+		SetEvent(evtDone);
 	}
 	env->CallStaticVoidMethod(m_dx9SageClass, m_dx9UpdateMethodID,
 		(jint) 0xCAFEBABE, 0, 0, 0, 0, 0, 0);
@@ -376,12 +382,9 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 	SAFE_RELEASE(pVideoTextureY);
 	SAFE_RELEASE(pVideoTextureU);
 	SAFE_RELEASE(pVideoTextureV);
-	if (myPtr !=NULL)
-		UnmapViewOfFile(myPtr);
-	if (evtReady != NULL)
-		CloseHandle(evtReady);
-	if (evtDone !=NULL)
-		CloseHandle(evtDone);
+	UnmapViewOfFile(myPtr);
+	CloseHandle(evtReady);
+	CloseHandle(evtDone);
 	CloseHandle(fileMap);
 	slog((env, "Async video renderer cleanup is complete.\r\n"));
 }
@@ -3105,7 +3108,7 @@ JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_getVideoSnapshot0
 	{
 // NOTE: WE SHOULD FIX THIS SO ITS NOT A PROBLEM
 		pSnapshotSurf->UnlockRect();
-		hr = E_FAIL;
+		hr = -1;
 		TEST_AND_BAIL
 	}
 	memcpy(nativeImageData, lockedRect.pBits, 4*videoWidth*videoHeight);
--- a/native/dll/DX93D/SageTVDX93D.cpp
+++ b/native/dll/DX93D/SageTVDX93D.cpp
@@ -3109,7 +3109,7 @@ JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_getVideoSnapshot0
 	{
 // NOTE: WE SHOULD FIX THIS SO ITS NOT A PROBLEM
 		pSnapshotSurf->UnlockRect();
-		hr = -1;
+		hr = E_FAIL;
 		TEST_AND_BAIL
 	}
 	memcpy(nativeImageData, lockedRect.pBits, 4*videoWidth*videoHeight);
--- a/native/dll/DX93D/SageTVDX93D.cpp
+++ b/native/dll/DX93D/SageTVDX93D.cpp
@@ -300,22 +300,25 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 		return;
 	}
 	const char* cName = env->GetStringUTFChars(sharedMemPrefix, NULL);
-	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE,
-		0, 1920*540*3 + 1024, cName);
+	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, 1920 * 540 * 3 + 1024, cName);
+	if (fileMap == NULL)
+		return;
 	char buf[256];
 	asyncRenderCount++;
 	Sleep(100); // wait more than 50 msec so if a prior MPlayer process was killed forcibly the loop below would have
 	// had time to exit and cleanup its pointers before we start on our work
-	strcpy(buf, cName);
-	strcat(buf, "FrameReady");
+	strcpy_s(buf, sizeof(buf), cName);
+	strcat_s(buf, sizeof(buf), "FrameReady");
 	HANDLE evtReady = CreateEvent(NULL, FALSE, FALSE, buf);
-	strcpy(buf, cName);
-	strcat(buf, "FrameDone");
+	strcpy_s(buf, sizeof(buf), cName);
+	strcat_s(buf, sizeof(buf), "FrameDone");
 	HANDLE evtDone = CreateEvent(NULL, FALSE, FALSE, buf);
 	env->ReleaseStringUTFChars(sharedMemPrefix, cName);
 	slog((env, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone));
 	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ|FILE_MAP_WRITE, 0, 0, 0);
 	unsigned int* myData = (unsigned int*) myPtr;
+	if (myData == NULL)
+		return;
 	slog((env, "Starting to read...0x%p\r\n", myPtr));
 	int configured = 0;
 	int j = 0;
@@ -325,13 +328,14 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 		j++;
 		// Decrease this wait time so that if there's case where MPlayer has to be forcibly quit we
 		// can exit this loop in a short amount of time
-		if (WAIT_OBJECT_0 != WaitForSingleObject(evtReady, 50))
-		{
-			if (env->GetLongField(jo, fid_pD3DObject) == 0)
-				break; // 3D system is dead
-			//printf("WAIT TIMEOUT EXPIRED!!!\r\n");
-			continue;
-		}
+		if (evtReady != NULL)
+			if (WAIT_OBJECT_0 != WaitForSingleObject(evtReady, 50))
+			{
+				if (env->GetLongField(jo, fid_pD3DObject) == 0)
+					break; // 3D system is dead
+				//printf("WAIT TIMEOUT EXPIRED!!!\r\n");
+				continue;
+			}
 		int currCmd = (myData[0] >> 24) & 0xFF;
 		//slog((env, "Got video cmd 0x%x\r\n", currCmd));
 		if (currCmd == 0x80)
@@ -363,7 +367,8 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 		else
 		{
 		}
-		SetEvent(evtDone);
+		if (evtDone != NULL)
+			SetEvent(evtDone);
 	}
 	env->CallStaticVoidMethod(m_dx9SageClass, m_dx9UpdateMethodID,
 		(jint) 0xCAFEBABE, 0, 0, 0, 0, 0, 0);
@@ -371,9 +376,12 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 	SAFE_RELEASE(pVideoTextureY);
 	SAFE_RELEASE(pVideoTextureU);
 	SAFE_RELEASE(pVideoTextureV);
-	UnmapViewOfFile(myPtr);
-	CloseHandle(evtReady);
-	CloseHandle(evtDone);
+	if (myPtr !=NULL)
+		UnmapViewOfFile(myPtr);
+	if (evtReady != NULL)
+		CloseHandle(evtReady);
+	if (evtDone !=NULL)
+		CloseHandle(evtDone);
 	CloseHandle(fileMap);
 	slog((env, "Async video renderer cleanup is complete.\r\n"));
 }
@@ -387,7 +395,9 @@ JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_initDX9SageRenderer0(J
 																		   jint width, jint height, jlong winID)
 {
 	JNI_TRY;
-	CoInitializeEx(NULL, COM_THREADING_MODE);
+	HRESULT hr = CoInitializeEx(NULL, COM_THREADING_MODE);
+	if (FAILED(hr))
+		return JNI_FALSE;
 	slog((env, "Initializing DirectX9\r\n"));
 	if (!fid_pD3DObject)
 	{
@@ -457,7 +467,7 @@ JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_initDX9SageRenderer0(J
 	}
 
 	D3DDISPLAYMODE d3ddm;
-	HRESULT hr = pD3D->GetAdapterDisplayMode(D3DADAPTER_DEFAULT, &d3ddm);
+	hr = pD3D->GetAdapterDisplayMode(D3DADAPTER_DEFAULT, &d3ddm);
 	TEST_AND_BAIL_NODEV
 	env->SetLongField(jo, fid_pD3DObject, (jlong) pD3D);
 
@@ -972,7 +982,7 @@ JNIEXPORT jlong JNICALL Java_sage_DirectX9SageRenderer_createD3DTextureFromMemor
 
     DWORD storedWidth = 1;
     DWORD storedHeight = 1;
-    if(requiresPow2Textures)
+    if (requiresPow2Textures)
     {
 //		slog((env, "Converting texture to power of 2 size\r\n"));
         while (storedWidth < ((DWORD) width))
@@ -1036,11 +1046,13 @@ JNIEXPORT jlong JNICALL Java_sage_DirectX9SageRenderer_createD3DTextureFromMemor
 	imgRect.bottom = storedHeight;
 	void* myMemory = env->GetDirectBufferAddress(data);
 	unsigned long* swapOrg = (unsigned long*)myMemory;
-	unsigned long* swapper = (unsigned long*)malloc(storedWidth*storedHeight*sizeof(unsigned long));
-	DWORD x,y;
+	unsigned long* swapper = (unsigned long*)calloc((storedWidth*storedHeight), sizeof(unsigned long));  // change from malloc to calloc to prevent 6386 warning
+	if (swapper == nullptr)
+		return 0;
+	DWORD x, y;
 	DWORD jheight = (DWORD) height;
 	DWORD jwidth = (DWORD) width;
-	for(y = 0; y < jheight; y++)
+	for (y = 0; y < jheight; y++)
 	{
 		int idxNew = y * storedWidth;
 		int idxOld = y * jwidth;
@@ -1065,6 +1077,7 @@ JNIEXPORT jlong JNICALL Java_sage_DirectX9SageRenderer_createD3DTextureFromMemor
 		0); // no color key
 
 	free(swapper);
+
 	TEST_AND_PRINT
 	if (hr != D3D_OK) 
 	{
@@ -3092,7 +3105,7 @@ JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_getVideoSnapshot0
 	{
 // NOTE: WE SHOULD FIX THIS SO ITS NOT A PROBLEM
 		pSnapshotSurf->UnlockRect();
-		hr = -1;
+		hr = E_FAIL;
 		TEST_AND_BAIL
 	}
 	memcpy(nativeImageData, lockedRect.pBits, 4*videoWidth*videoHeight);
--- a/native/dll/DX93D/SageTVDX93D.cpp
+++ b/native/dll/DX93D/SageTVDX93D.cpp
@@ -300,7 +300,8 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
 		return;
 	}
 	const char* cName = env->GetStringUTFChars(sharedMemPrefix, NULL);
-	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, 1920 * 540 * 3 + 1024, cName);
+	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE,
+		0, 1920*540*3 + 1024, cName);
 	if (fileMap == NULL)
 		return;
 	char buf[256];
--- a/native/dll/Win32DLL/OpenGLVideoRenderer.cpp
+++ b/native/dll/Win32DLL/OpenGLVideoRenderer.cpp
@@ -47,9 +47,9 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
 	glVideoServerActive = 1;
 	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE,
 		0, 1920*540*3 + 1024, shmemPrefix);
-	char buf[256];
 	if (fileMap == NULL)
 		return JNI_FALSE;
+	char buf[256];
 	strcpy_s(buf, sizeof(buf), shmemPrefix);
 	strcat_s(buf, sizeof(buf), "FrameReady");
 	HANDLE evtReady = CreateEvent(NULL, FALSE, FALSE, buf);
--- a/native/dll/Win32DLL/OpenGLVideoRenderer.cpp
+++ b/native/dll/Win32DLL/OpenGLVideoRenderer.cpp
@@ -46,24 +46,33 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
 	shMemCounter++;
 	glVideoServerActive = 1;
 	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, 1920*540*3 + 1024, shmemPrefix);
-	if (fileMap == NULL)  // Should never be NULL
-		return JNI_FALSE;
 	char buf[256];
 	strcpy_s(buf, sizeof(buf), shmemPrefix);
 	strcat_s(buf, sizeof(buf), "FrameReady");
 	HANDLE evtReady = CreateEvent(NULL, FALSE, FALSE, buf);
-	if (evtReady == NULL)  // Should never be NULL
-		return JNI_FALSE;
 	strcpy_s(buf, sizeof(buf), shmemPrefix);
 	strcat_s(buf, sizeof(buf), "FrameDone");
 	HANDLE evtDone = CreateEvent(NULL, FALSE, FALSE, buf);
-	if (evtDone == NULL)  // Should never be NULL
+	if (fileMap == NULL || evtDone == NULL || evtReady == NULL)
+	{
+		if (evtReady != NULL)
+			CloseHandle(evtReady);
+		if (evtDone != NULL)
+			CloseHandle(evtDone);
+		if (fileMap != NULL)
+			CloseHandle(fileMap);
 		return JNI_FALSE;
+	}
 	fprintf(stderr, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone);
 	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ | FILE_MAP_WRITE, 0, 0, 0);
 	unsigned int* myData = (unsigned int*) myPtr;
-	if (myData == NULL)  // Should never be NULL
+	if (myData == NULL || myPtr != NULL)
+	{
+		if (myPtr != NULL)
+			UnmapViewOfFile(myPtr);
 		return JNI_FALSE;
+	}
+
 	jobject byteBuffer = env->NewDirectByteBuffer(myPtr + 1024, 1920*540*3);
 	if (env->ExceptionOccurred())
 		return JNI_FALSE; // let the exception propagate
@@ -99,7 +108,6 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
 			env->CallBooleanMethod(jo, m_glUpdateMethodID, myData[1], byteBuffer);
 			if (env->ExceptionOccurred())
 				return JNI_FALSE; // let the exception propagate
-			if (evtDone !=NULL)
 				SetEvent(evtDone);
 		}
 		else if (currCmd == 0x82)
@@ -111,14 +119,10 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
 				ResetEvent(evtDone);
 		}
 	}
-	if (myPtr !=NULL)
-		UnmapViewOfFile(myPtr);
-	if (evtReady != NULL)
-		CloseHandle(evtReady);
-	if (evtDone != NULL)
-		CloseHandle(evtDone);
-	if (fileMap != NULL)
-		CloseHandle(fileMap);
+	UnmapViewOfFile(myPtr);
+	CloseHandle(evtReady);
+	CloseHandle(evtDone);
+	CloseHandle(fileMap);
 	return JNI_TRUE;
 }
 
--- a/native/dll/Win32DLL/OpenGLVideoRenderer.cpp
+++ b/native/dll/Win32DLL/OpenGLVideoRenderer.cpp
@@ -45,7 +45,8 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
 	sprintf_s(shmemPrefix, sizeof(shmemPrefix), "SageTV-%d-%d", GetCurrentProcessId(), shMemCounter);
 	shMemCounter++;
 	glVideoServerActive = 1;
-	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, 1920*540*3 + 1024, shmemPrefix);
+	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE,
+		0, 1920*540*3 + 1024, shmemPrefix);
 	char buf[256];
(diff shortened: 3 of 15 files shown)


native/dll/DX93D/SageTVDX93D.cpp
```diff
@@ -300,22 +300,25 @@ JNIEXPORT void JNICALL Java_sage_DirectX9SageRenderer_asyncVideoRender0
...
-	strcpy(buf, cName);
-	strcat(buf, "FrameDone");
+	strcpy_s(buf, sizeof(buf), cName);
+	strcat_s(buf, sizeof(buf), "FrameDone");
 	HANDLE evtDone = CreateEvent(NULL, FALSE, FALSE, buf);
 	env->ReleaseStringUTFChars(sharedMemPrefix, cName);
 	slog((env, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone));
 	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ|FILE_MAP_WRITE, 0, 0, 0);
 	unsigned int* myData = (unsigned int*) myPtr;
+	if (myData == NULL)
```
Narflex: I'd change this to be 'if (myPtr == NULL || evtReady == NULL || evtDone == NULL)' and then put a block below like at the end of the file where you release all of the non-null objects so they get properly cleaned up in case of error. Then get rid of your other NULL checks below.

native/dll/DX93D/SageTVDX93D.cpp
```diff
@@ -3092,7 +3105,7 @@ JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_getVideoSnapshot0
 	{
 // NOTE: WE SHOULD FIX THIS SO ITS NOT A PROBLEM
 		pSnapshotSurf->UnlockRect();
-		hr = -1;
+		hr = E_FAIL;
```
Narflex: Why are you changing this?

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
```diff
@@ -32,34 +32,41 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
...
+	strcpy_s(buf, sizeof(buf), shmemPrefix);
+	strcat_s(buf, sizeof(buf), "FrameDone");
 	HANDLE evtDone = CreateEvent(NULL, FALSE, FALSE, buf);
+	if (evtDone == NULL)  // Should never be NULL
+		return JNI_FALSE;
 	fprintf(stderr, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone);
-	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ|FILE_MAP_WRITE, 0, 0, 0);
+	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ | FILE_MAP_WRITE, 0, 0, 0);
 	unsigned int* myData = (unsigned int*) myPtr;
+	if (myData == NULL)  // Should never be NULL
```
Narflex: Do the same thing as in the DX9 renderer where you check them all here and then cleanup anything that did work and exit on failure.

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -202,23 +202,24 @@ JNIEXPORT jboolean JNICALL Java_sage_Sage_writeStringValue(JNIEnv *env,
...
 	jboolean rv = JNI_FALSE;
 	if (root == sage_Sage_HKEY_CLASSES_ROOT) rootKey = HKEY_CLASSES_ROOT;
 	else if (root == sage_Sage_HKEY_CURRENT_CONFIG) rootKey = HKEY_CURRENT_CONFIG;
 	else if (root == sage_Sage_HKEY_CURRENT_USER) rootKey = HKEY_CURRENT_USER;
 	else if (root == sage_Sage_HKEY_LOCAL_MACHINE) rootKey = HKEY_LOCAL_MACHINE;
 	else if (root == sage_Sage_HKEY_USERS) rootKey = HKEY_USERS;
-
-	if (rootKey && RegCreateKeyEx(rootKey, keyString, 0, 0, REG_OPTION_NON_VOLATILE, KEY_ALL_ACCESS,
-		0, &myKey, 0) == ERROR_SUCCESS)
+	if (rootKey != NULL)
```
Narflex: You don't need this line because rootKey is checked in the conditional 2 lines down.

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -245,17 +246,15 @@ jboolean JNICALL Java_sage_Sage_writeDwordValue(JNIEnv *env,
 	else if (root == sage_Sage_HKEY_CURRENT_USER) rootKey = HKEY_CURRENT_USER;
 	else if (root == sage_Sage_HKEY_LOCAL_MACHINE) rootKey = HKEY_LOCAL_MACHINE;
 	else if (root == sage_Sage_HKEY_USERS) rootKey = HKEY_USERS;
-
-	if (rootKey && RegCreateKeyEx(rootKey, keyString, 0, 0, REG_OPTION_NON_VOLATILE, KEY_ALL_ACCESS,
-		0, &myKey, 0) == ERROR_SUCCESS)
-	{
-		if (RegSetValueEx(myKey, valueNameString, 0, REG_DWORD, (LPBYTE) &dvalue,
-			sizeof(dvalue)) == ERROR_SUCCESS)
+	if (rootKey != NULL)
```
Narflex: Remove this line

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -272,19 +271,20 @@ jboolean JNICALL Java_sage_Sage_removeRegistryValue(JNIEnv *env,
...
 	else if (root == sage_Sage_HKEY_CURRENT_USER) rootKey = HKEY_CURRENT_USER;
 	else if (root == sage_Sage_HKEY_LOCAL_MACHINE) rootKey = HKEY_LOCAL_MACHINE;
 	else if (root == sage_Sage_HKEY_USERS) rootKey = HKEY_USERS;
-
-	if (rootKey && RegOpenKeyEx(rootKey, keyString, 0, KEY_ALL_ACCESS, &myKey) == ERROR_SUCCESS)
-	{
-		rv = (RegDeleteValue(myKey, valueNameString) == ERROR_SUCCESS);
-		RegCloseKey(myKey);
-	}
+	if (rootKey != NULL)
```
Narflex: Remove this line

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -399,8 +397,7 @@ JNIEXPORT jobjectArray JNICALL Java_sage_Sage_getRegistrySubkeys(JNIEnv *env, jc
 	
 	RegCloseKey(myKey);
 	env->ReleaseStringUTFChars(key, keyString);
-	return (jobjectArray) env->CallObjectMethod(vec, vecToArrayMeth,
-		env->NewObjectArray(env->CallIntMethod(vec, vecSizeMeth), strClass, NULL));
+	return (jobjectArray) env->CallObjectMethod(vec, vecToArrayMeth, env->NewObjectArray(env->CallIntMethod(vec, vecSizeMeth), strClass, NULL));
```
Narflex: There's a bunch of places in this file where you put multiple lines onto single lines...please revert those because you're going to end up with lines too long for the style that was used in this file.

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -572,7 +567,7 @@ JNIEXPORT void JNICALL Java_sage_Sage_addTaskbarIcon0(JNIEnv *env, jclass jc, jl
 	iconData.hIcon = LoadIcon(GetModuleHandle(NULL), MAKEINTRESOURCE(iconID));
 	static jclass sagetvClass = (jclass) env->NewGlobalRef(env->FindClass("sage/SageTV"));
 	env->ExceptionClear();
-	strcpy(iconData.szTip, sagetvClass ? "SageTV 2" : "SageRecorder");
+	strcpy_s(iconData.szTip, sizeof(iconData.szTip), sagetvClass ? "SageTV 2" : "SageRecorder");
```
Narflex: Just make the 'SageTV 2' be 'SageTV'...that clearly hasn't been updated since 2004. :)

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -709,7 +703,9 @@ JNIEXPORT jlong JNICALL Java_sage_Sage_getEventTime0(JNIEnv *env, jclass jc)
 //    }
 	// NARFLEX - I have no idea why this was using GetMessageTime; but that causes the value to
 	// vary between threads which is very, very bad so we don't use that at all anymore.
+#pragma warning(disable:28159) // Static Code Analysis: When we stop supporting XP we can use GetTickCount64()
```
Narflex: Can't you just put this once at the top of the file? Putting it 4 times in the same function looks really ugly.

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -928,12 +921,12 @@ JNIEXPORT jlong JNICALL Java_sage_UIManager_findWindow(
 JNIEXPORT jboolean JNICALL Java_sage_DirectX9SageRenderer_hasDirectX90
   (JNIEnv *env, jclass jc)
 {
-    TCHAR szPath[512];
-    TCHAR szFile[512];
-    if (GetSystemDirectory(szPath, 512) != 0)
+    TCHAR szPath[_MAX_PATH];
+    TCHAR szFile[_MAX_PATH];
+    if (GetSystemDirectory(szPath, _MAX_PATH) != 0)
```
Narflex: Use sizeof(szPath) instead of _MAX_PATH here.

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -1013,7 +1011,7 @@ void loadAWTLib()
 		*goodSlash = 0;
 		goodSlash = strrchr(jvmPath, '\\');
 		if (!goodSlash) return;
-		strcpy(goodSlash + 1, "jawt.dll");
+		strcpy_s(goodSlash + 1, sizeof(goodSlash) + 1, "jawt.dll");
```
Narflex: You can't use sizeof(goodslash) because it'll give you the pointer size. What you can do instead is:
*(goodSlash +1) = 0;
strncat_s(jvmPath, sizeof(jvmPath), "jawt.dll");

native/dll/Win32DLL/SageTVWin32DLL.def
```diff
@@ -3,4 +3,3 @@
 LIBRARY      "SageTVWin32"
 
 SECTIONS
```
Narflex: I don't know what changing this does...so unless you can explain to me why you're changing it, please revert this part of the change.

native/dll/Win32DLL/WMATagParser.cpp
```diff
@@ -148,17 +148,17 @@ class CWMProp : public IWMReaderCallback
 			switch(attType)
 			{
 			case WMT_TYPE_QWORD:
-				_ui64tow( *((QWORD*)pbValue), attValueName, 10);
+				_ui64tow_s( *((QWORD*)pbValue), attValueName, sizeof(attValueName), 10);
```
Narflex: You can't use sizeof(attValueName). Just set a variable to be cbValue + 64 and reuse that for the allocation and in these calls.

native/dll/Win32DLL/WindowsServiceControl.cpp
```diff
@@ -651,47 +647,37 @@ GetAccountSid(
...
-            // 
-            // reallocate memory
-            // 
-            if((*Sid=HeapReAlloc(
-                        GetProcessHeap(),
-                        0,
-                        *Sid,
-                        cbSid
-                        )) == NULL) __leave;
+	if ((Sid != nullptr) && (ReferencedDomain != NULL))
```
Narflex: You don't need to check Sid or ReferencedDomain because they are already being checked above and the __leave is done to terminate the section.

native/dll/Win32DLL/WindowsServiceControl.cpp
```diff
@@ -651,47 +647,37 @@ GetAccountSid(
...
+		&peUse
+	))
+	{
+		if (GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+		{
+			// 
+			// reallocate memory
+			//
+
+#pragma warning(disable:6308) // Static Code Analysis" 'realloc' might return null pointer, will cause the original memory block to be leaked..
```
Narflex: This sounds like something that should be fixed rather than hidden with a pragma warning

native/dll/WinRawInput/WinRawInput.def
```diff
@@ -3,4 +3,3 @@
 LIBRARY      "WinRawInput"
 
 SECTIONS
-	.shared READ WRITE SHARED
```
Narflex: Again, don't change this unless you can explain why

native/dll/DX93D/SageTVDX93D.cpp
JREkiwi: OK.  Done

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done  + the other one too :)

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: You're right it is ugly. I was trying not to globally disable warnings, but this one really could/should be disabled at the top of the file.
Done

native/dll/Win32DLL/WMATagParser.cpp
JREkiwi: I'll get these sorted in my head eventually :)
Done

native/dll/Win32DLL/SageTVWin32DLL.def
JREkiwi: When it's there, we get a compile warning of 
`LINK : warning LNK4039: section '.shared' specified with /SECTION option does not exist`
which from my reading is because it's defined in the .def file but not used

native/common/jni-util.cpp
```diff
@@ -692,7 +690,11 @@ jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, cha
 	}
 	const char* cstr = env->GetStringUTFChars(jstr, NULL);
 	*rvCount = (unsigned long) MIN(rvlen, strlen(cstr));
+#ifdef WIN32
```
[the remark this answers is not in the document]
JREkiwi: Done, I think? :)

native/dll/DX93D/SageTVDX93D.cpp
JREkiwi: Static Analysis gives warning
`warning C6226: Implicit cast between semantically different integer types:  assigning -1 to HRESULT.  Consider using E_FAIL instead.`

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done, is the style documented somewhere?

native/dll/Win32DLL/WindowsServiceControl.cpp
JREkiwi: From what I could tell it's an issue with the Static Code Analysis.
It is checking for a NULL return and we already check that the arguments passed to it are not NULL

native/dll/Win32DLL/WindowsServiceControl.cpp
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
Narflex: No, we don't have documented style for this. There's LOTS of style inconsistency in the SageTV codebase, so for new code we can make it look better...but going around changing a bunch of existing code that's not modified for other reasons just makes it harder to track down changes that introduced problems in those same areas.

native/dll/Win32DLL/SageTVWin32DLL.def
Narflex: OK

native/dll/Win32DLL/WindowsServiceControl.cpp
Narflex: I looked into this, and what it's referring to is that if the HeapReAlloc function fails, then it will return NULL. In that case, the Sid block we allocated before will still have been allocated...BUT since we set the return value to that, we have then lost that value and cannot deallocate it. So this is a legitimate warning.

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: I was having difficulty following the code, so had changed them for readability.
Have reverted them now

native/dll/Win32DLL/WindowsServiceControl.cpp
JREkiwi: I had a few goes at this before, without success, but I think I have  it now.

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: I'm struggling with this one.  I don't understand how it works now.
We get a ptr with the last \ (Understand that)
We find the 2nd to last  \ (Don't understand)
How do we end up with the required resultant string in jvmPath?
initial jvmPath = C:\Program Files\Java\jre1.8.0_241\bin\server\jvm.dll
jvmPath in sageLoadedAwtLib = LoadLibrary(jvmPath); = C:\Program Files\Java\jre1.8.0_241\bin\jawt.dll

JREkiwi: Very slowly getting there `This pointer (plus one) will point to the base name of the file`
Still trying to figure out how that becomes the new value on jvmPath

Narflex: OK, so here's what the current code does:

1. Find the last backslash and get the pointer to it:
char* goodSlash = strrchr(jvmPath, '\\');
2. If we didn't find it, return.
if (!goodSlash) return;
3. Replace the character at the last slash with the null terminator. The essentially truncates the string so the last character is now the one before the original last slash (i.e. it removes \jvm.dll)
*goodSlash = 0;
4. Find the last slash in the string again. This will be the second one originally, but now that is the last one.
goodSlash = strrchr(jvmPath, '\\');

Then at that point we can copy into that position after the slash (original code). Or what I'm suggesting is to then truncate that string again after that slash (i.e. it would end with bin\) and then do a strcat_s on it since we know the size of jvmPath easily. (you could also do other pointer math to get a proper value for strcpy_s, but that's easier to get wrong)

Make sense now? :)

JREkiwi: Thanks for spending the time walking me through this 👍 it's making more sense. When it it's doing this, is it changing jvmPath too?

JREkiwi: Thanks for spending the time walking me through this 👍 it's making more sense. When it it's doing this, is it changing jvmPath too?

Narflex: Yes, it's changing jvmPath. When that first strrchr is called, it's setting goodSlash to be a pointer into the jvmPath array. So if you modify goodSlash, you are actually modifying jvmPath since it's pointing inside of its array.

JREkiwi: Ahhh. Ok. I'll finish those changes when I get home
Thanks once more :)

native/dll/Win32DLL/WindowsServiceControl.cpp
JREkiwi: Still battling `Using uninitialized memory 'ReferencedDomain'` on this

wnjj: What line does it flag that error on?

For the memory leak, consider changing the HeapRealloc to use a new variable which you copy into the original only upon success.  Ditto for ReferencedDomain.  You probably don’t need the “else” I added there if ”__leave” returns.  I didn’t find its definition yet.

PSID newSid;
if ((newSid = HeapReAlloc(GetProcessHeap(), 0, *Sid, cbSid)) == nullptr)
  __leave;
else
  *Sid = newSid;

JREkiwi: I'd got to here`			//

			PSID *tmpSid = new PSID;
			*tmpSid = *Sid;
			if ((*tmpSid = HeapReAlloc(GetProcessHeap(), 0, *Sid, cbSid)) == nullptr)
			{
				HeapFree(GetProcessHeap(), 0, *tmpSid);
				*tmpSid = NULL;
				__leave;
			}
			else
			{
				*Sid = tmpSid;
				HeapFree(GetProcessHeap(), 0, *tmpSid);
				*tmpSid = NULL;
			}

			LPTSTR tmpReferencedDomain = nullptr;
			if ((tmpReferencedDomain = (LPTSTR)HeapReAlloc(GetProcessHeap(), 0, ReferencedDomain, cchReferencedDomain * sizeof(TCHAR))) == NULL)
			{
				HeapFree(GetProcessHeap(), 0, tmpReferencedDomain);
				__leave;
			}
			else
			{
				strcpy_s(ReferencedDomain, strlen(ReferencedDomain), tmpReferencedDomain);
				HeapFree(GetProcessHeap(), 0, tmpReferencedDomain);
			}`
and I get the "using Uninitialized memory" warning on `strcpy_s(ReferencedDomain, strlen(ReferencedDomain), tmpReferencedDomain);` if that is `ReferencedDomain = tmpReferencedDomain;` then I get the "using Uninitialized memory" warning down @ 
`    // 
    // Cleanup and indicate failure, if appropriate.
    // 

    HeapFree(GetProcessHeap(), 0, ReferencedDomain);
`

Narflex: A few comments on the above inlined below...

> I'd got to here` //
> 
> ```
> 		PSID *tmpSid = new PSID;
Should instead be 'PSID tmpSid = NULL;', you shouldn't be allocating anything here because this is just temporary storage for a return value. It's also already a pointer (that's what the P prefix means) so you don't need to make it a double pointer (we need to have a double pointer passed into the function because we are using that to return the allocated structure).
> 		*tmpSid = *Sid;
If you did this, and had 'new PSID' above, you would be leaking that allocated PSID object. You don't need to do this anyways because tmpSid is getting assigned on the next line, so any assignment here won't matter at all, just delete this line.
> 		if ((*tmpSid = HeapReAlloc(GetProcessHeap(), 0, *Sid, cbSid)) == nullptr)
Get rid of the asterisk on 'tmpSid' here.
> 		{
> 			HeapFree(GetProcessHeap(), 0, *tmpSid);
This is trying to free the wrong thing. tmpSid will be null, because that's what the conditional checks above. Instead, here you should be freeing '*Sid' and then setting it to NULL on the next line. 
> 			*tmpSid = NULL;
> 			__leave;
> 		}
> 		else
> 		{
> 			*Sid = tmpSid;
> 			HeapFree(GetProcessHeap(), 0, *tmpSid);
Get rid of this line because otherwise you are freeing what you just re-allocated because you set them equal on the above line.
> 			*tmpSid = NULL;
This line is not necessary, but it doesn't hurt either.
> 		}
> 
> 		LPTSTR tmpReferencedDomain = nullptr;
> 		if ((tmpReferencedDomain = (LPTSTR)HeapReAlloc(GetProcessHeap(), 0, ReferencedDomain, cchReferencedDomain * sizeof(TCHAR))) == NULL)
> 		{
> 			HeapFree(GetProcessHeap(), 0, tmpReferencedDomain);
You should be freeing ReferencedDomain here instead since that's the initially allocated thing you'd be leaking if the realloc fails.
> 			__leave;
> 		}
> 		else
> 		{
> 			strcpy_s(ReferencedDomain, strlen(ReferencedDomain), tmpReferencedDomain);
You shouldn't be trying to copy a string here. You should just be putting the allocated memory on the right pointer. So instead this should just be: 'ReferencedDomain = tmpReferencedDomain;'
> 			HeapFree(GetProcessHeap(), 0, tmpReferencedDomain);
Get rid of this line, otherwise you are freeing the memory you just allocated.
> 		}`
> ```
> 
> and I get the "using Uninitialized memory" warning on `strcpy_s(ReferencedDomain, strlen(ReferencedDomain), tmpReferencedDomain);` if that is `ReferencedDomain = tmpReferencedDomain;` then I get the "using Uninitialized memory" warning down @
> ` //
> // Cleanup and indicate failure, if appropriate.
> //
> 
> ```
> HeapFree(GetProcessHeap(), 0, ReferencedDomain);
Put an 'if (ReferencedDomain)' (i.e. not null, that's the shortcut way of doing it) above this and I think that'll clean up the error. It's possible ReferencedDomain is never allocated if the original allocation of Sid fails and that's likely what it is warning you on.
> ```
> 
> `

JREkiwi: Is this a record?  Looks like I've succeeded in getting pretty much every line wrong :)

JREkiwi: > It's possible ReferencedDomain is never allocated if the original allocation of Sid fails and that's likely what it is warning you on.

Yes, I can see that this is what's happening, so I assume we'll need to initialize ReferencedDomain up here somehow
 
`BOOL GetAccountSid(LPTSTR SystemName, LPTSTR AccountName, PSID *Sid)

{
    LPTSTR ReferencedDomain = nullptr;
    DWORD cbSid=128;    // initial allocation attempt
`

JREkiwi: I don't see how we ever end up at the HeapRealloc if we leave the __try block when HeapAlloc fails

Narflex: You shouldn't hit HeapRealloc if the HeapAlloc fails and you _leave the try block. What error are you seeing now and what does the code look like?

JREkiwi: If I put an 'if (ReferencedDomain), I get a warning as ReferencedDomain can be uninitialized if the HeapAlloc fails.

Doesn't the HeapAlloc have to fail in order to get ERROR_INSUFFICIENT_BUFFER returned from GetLastError(), but we'll never get that becuase we leave the __try?

Narflex: Can you post the code you have for that whole function please?

JREkiwi: Here you go

```
BOOL GetAccountSid(LPTSTR SystemName, LPTSTR AccountName, PSID *Sid)
{
    LPTSTR ReferencedDomain = NULL;
    DWORD cbSid=128;    // initial allocation attempt
    DWORD cchReferencedDomain=16; // initial allocation size
    SID_NAME_USE peUse;
    BOOL bSuccess=FALSE; // assume this function will fail

   __try {

    // 
    // initial memory allocations
    // 
    if((*Sid=HeapAlloc(GetProcessHeap(), 0, cbSid)) == NULL) __leave;

    if((ReferencedDomain=(LPTSTR)HeapAlloc(GetProcessHeap(), 0, cchReferencedDomain * sizeof(TCHAR))) == NULL) __leave;

    // 
    // Obtain the SID of the specified account on the specified system.
    // 
	while (!LookupAccountName(
	SystemName,         // machine to lookup account on
	AccountName,        // account to lookup
	*Sid,               // SID of interest
	&cbSid,             // size of SID
	ReferencedDomain,   // domain account was found on
	&cchReferencedDomain,
	&peUse
	))
	{
		if (GetLastError() == ERROR_INSUFFICIENT_BUFFER)
		{
			// 
			// reallocate memory
			//

			PSID tmpSid = NULL;
			if ((tmpSid = HeapReAlloc(GetProcessHeap(), 0, *Sid, cbSid)) == NULL)
			{
				HeapFree(GetProcessHeap(), 0, *Sid);
				*Sid = NULL;
				__leave;
			}
			else
			{
				*Sid = tmpSid;
			}

			LPTSTR tmpReferencedDomain = NULL;
			if ((tmpReferencedDomain = (LPTSTR)HeapReAlloc(GetProcessHeap(), 0, ReferencedDomain, cchReferencedDomain * sizeof(TCHAR))) == NULL)
			{
				HeapFree(GetProcessHeap(), 0, ReferencedDomain);
				__leave;
			}
			else
			{
				ReferencedDomain = tmpReferencedDomain;
			}
        }
        else __leave;
    }

    // 
    // Indicate success.
    // 
    bSuccess=TRUE;

    } // finally
    __finally 
	{

    // 
    // Cleanup and indicate failure, if appropriate.
    // 

	if (ReferencedDomain)
		HeapFree(GetProcessHeap(), 0, ReferencedDomain);

    if(!bSuccess)
	{
        if(*Sid)
		{
            HeapFree(GetProcessHeap(), 0, *Sid);
            *Sid = NULL;
        }
	}

    } // finally

    return bSuccess;
}

```

JREkiwi: OK, I can now see that if the initial HeapAlloc is good and LookupAccountName fails with ERROR_INSUFFICIENT_BUFFER it will do the HeapRealloc

JREkiwi: I think my only problem now is at
 `	if (ReferencedDomain)`
`		HeapFree(GetProcessHeap(), 0, ReferencedDomain);`
I'm getting
`Using uninitialized memory 'ReferencedDomain'` @  `	if (ReferencedDomain)`

wnjj: I think you missed setting ReferencedDomain to NULL when you free it above:
```
			{
				HeapFree(GetProcessHeap(), 0, ReferencedDomain);
				__leave;
			}
```

If that executes, the finally block will try to free it again.  Try this:
```
			{
				HeapFree(GetProcessHeap(), 0, ReferencedDomain);
				ReferencedDomain = NULL;
				__leave;
			}
```

JREkiwi: That's it :)
Thanks

JREkiwi:
Looks like some of the indentation got screwed up

third_party/Microsoft/BaseClasses/wxutil.cpp
```diff
@@ -607,7 +610,8 @@ DWORD WINAPI WaitDispatchingMessages(
 
     HANDLE hObjects[2] = { hObject, hEvent };
     if (dwWait != INFINITE && dwWait != 0) {
-        dwStart = GetTickCount();
+		dwStart = GetTickCount();
```
Narflex: Fix this indentation please.

third_party/Microsoft/BaseClasses/wxutil.cpp
```diff
@@ -643,7 +647,7 @@ DWORD WINAPI WaitDispatchingMessages(
             PeekMessage(&msg, NULL, 0, 0, PM_NOREMOVE);
 
             if (dwWait != INFINITE && dwWait != 0) {
-                DWORD dwNow = GetTickCount();
+					DWORD dwNow = GetTickCount();
```
Narflex: Fix indentation please.

JREkiwi: Already Done

third_party/Microsoft/BaseClasses/wxutil.cpp
JREkiwi: Already done

Narflex:
And in the future...please limit your pull requests to much smaller changes. For stuff like this, ideally it'd be one file at a time (unless it's just a couple line change). It's too easy for me to start to glaze over when doing a review like this and miss something.

native/exe/MiniClientLauncher/MiniClientLauncher.cpp
```diff
@@ -161,21 +165,29 @@ int APIENTRY WinMain(HINSTANCE hInstance,
...
-	if(hLib == NULL) 
+
+	if (hLib == NULL)
 	{
-		// Failed, load the JVM from the registry instead
+		// Failed to find JRE in SageTV directory, load the JVM from the registry instead, by using the Windows Registry to locate the current version to use.
+
+		HKEY rootKey = HKEY_LOCAL_MACHINE;
+		char currVer[16];
+		HKEY myKey;  // myKey to NULL for RegQueryValueEx
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -202,6 +204,7 @@ JNIEXPORT jboolean JNICALL Java_sage_Sage_writeStringValue(JNIEnv *env,
 	const char* valueNameString = env->GetStringUTFChars(valueName, 0);
 	const char* valueString = env->GetStringUTFChars(value, 0);
 	HKEY rootKey;
+	rootKey = NULL;  // initialize to resolve C6001
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
```diff
@@ -91,15 +108,16 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
...
 			SetEvent(evtDone);
 		}
 		else if (currCmd == 0x82)
 		{
 			env->CallBooleanMethod(jo, m_glCloseMethodID);
 			if (env->ExceptionOccurred())
-				return JNI_FALSE; // let the exception propogate
-			ResetEvent(evtDone);
+				return JNI_FALSE; // let the exception propagate
+			if (evtDone != NULL)
```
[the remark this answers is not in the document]
JREkiwi: Already done

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
```diff
@@ -32,34 +32,51 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
...
-	strcat(buf, "FrameReady");
+	strcpy_s(buf, sizeof(buf), shmemPrefix);
+	strcat_s(buf, sizeof(buf), "FrameReady");
 	HANDLE evtReady = CreateEvent(NULL, FALSE, FALSE, buf);
-	strcpy(buf, shmemPrefix);
-	strcat(buf, "FrameDone");
+	strcpy_s(buf, sizeof(buf), shmemPrefix);
+	strcat_s(buf, sizeof(buf), "FrameDone");
 	HANDLE evtDone = CreateEvent(NULL, FALSE, FALSE, buf);
+	if (fileMap == NULL || evtDone == NULL || evtReady == NULL)
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
```diff
@@ -32,34 +32,51 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
...
+		return JNI_FALSE; // let the exception propagate
+	sprintf_s(shmemPrefix, sizeof(shmemPrefix), "SageTV-%d-%d", GetCurrentProcessId(), shMemCounter);
 	shMemCounter++;
 	glVideoServerActive = 1;
 	HANDLE fileMap = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE,
 		0, 1920*540*3 + 1024, shmemPrefix);
 	char buf[256];
-	strcpy(buf, shmemPrefix);
-	strcat(buf, "FrameReady");
+	strcpy_s(buf, sizeof(buf), shmemPrefix);
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/WindowsServiceControl.cpp
```diff
@@ -679,19 +685,30 @@ GetAccountSid(
...
-                        GetProcessHeap(),
-                        0,
-                        ReferencedDomain,
-                        cchReferencedDomain * sizeof(TCHAR)
-                        )) == NULL) __leave;
+
+            PSID tmpSid = NULL;
+            if ((tmpSid = HeapReAlloc(GetProcessHeap(), 0, *Sid, cbSid)) == NULL)
+            {
+                HeapFree(GetProcessHeap(), 0, *Sid);
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/WindowsServiceControl.cpp
```diff
@@ -679,19 +685,30 @@ GetAccountSid(
...
+            }
+            else
+            {
+            	*Sid = tmpSid;
+            }
+
+            LPTSTR tmpReferencedDomain = NULL;
+            if ((tmpReferencedDomain = (LPTSTR)HeapReAlloc(GetProcessHeap(), 0, ReferencedDomain, cchReferencedDomain * sizeof(TCHAR))) == NULL)
+            {
+                HeapFree(GetProcessHeap(), 0, ReferencedDomain);
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
```diff
@@ -32,34 +32,51 @@ JNIEXPORT jint JNICALL Java_sage_miniclient_OpenGLVideoRenderer_initVideoServer
...
+		if (evtDone != NULL)
+			CloseHandle(evtDone);
+		if (fileMap != NULL)
+			CloseHandle(fileMap);
+		return JNI_FALSE;
+	}
 	fprintf(stderr, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone);
 	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ|FILE_MAP_WRITE, 0, 0, 0);
 	unsigned int* myData = (unsigned int*) myPtr;
+	if (myData == NULL || myPtr != NULL)
```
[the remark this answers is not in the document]
JREkiwi: But doesn't the myPtr NULL check need to be after `unsigned char* myPtr = (unsigned char*)MapViewOfFile` @ 68
And the other NULL checks need to be before `fprintf(stderr, "Created FileMap` @ 67 ?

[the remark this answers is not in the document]
JREkiwi: How about doing this
```
    if (evtReady == NULL || evtDone == NULL)
	{
		if (evtReady != NULL)
			CloseHandle(evtReady);
		if (evtDone != NULL)
			CloseHandle(evtDone);
		if (fileMap != NULL)
			CloseHandle(fileMap);
		return JNI_FALSE;
	}
	fprintf(stderr, "Created FileMap=0x%p evtReady=0x%p evtDone=0x%p\r\n", fileMap, evtReady, evtDone);
	unsigned char* myPtr = (unsigned char*)MapViewOfFile(fileMap, FILE_MAP_READ|FILE_MAP_WRITE, 0, 0, 0);
    if (myPtr == NULL)
        return JNI_FALSE;
    unsigned int* myData = (unsigned int*) myPtr;

```

native/dll/Win32DLL/WindowsServiceControl.cpp
JREkiwi: It's only a problem because I have the HeapFree there. With that gone it is all OK.
As Jeff has pointed out, I can leave those to the later cleanup :)

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -1231,6 +1242,7 @@ JNIEXPORT jboolean JNICALL Java_sage_Sage_setupSystemHooks0
  	if ( WAIT_OBJECT_0 != dwWaitResult )
 	{
 		slog((env, "setup system shell hook failed (object is locked)\r\n"));
+		ReleaseMutex(hMutexMessaging);
```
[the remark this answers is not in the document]
JREkiwi: As CreateMutex did not return NULL, don't we need to release ownership of the mutex before Closing the handle ?

native/dll/Win32DLL/WMATagParser.cpp
```diff
@@ -145,20 +145,21 @@ class CWMProp : public IWMReaderCallback
 
 			// Display the attribute depending upon type.
 			WCHAR* attValueName = new WCHAR[cbValue + 64];
+			size_t aVNsize = (cbValue + 64);
```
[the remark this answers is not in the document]
JREkiwi: Done

JREkiwi:
> And in the future...please limit your pull requests to much smaller changes. For stuff like this, ideally it'd be one file at a time (unless it's just a couple line change). It's too easy for me to start to glaze over when doing a review like this and miss something.

Yeah, sorry about that, and noted. I started out trying to just do the MiniClientLauncher with the changes that had been done for the ServiceControlLaunch and what I thought was going to be just a couple of lines in some other files,  but then I got trapped in SageTVWin32DLL.cpp :)

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -952,33 +958,38 @@ void loadAWTLib()
...
 			return;
 		/*
-		 * Explicitly load jawt.dll by using the Windows Registry to locate the current version to use.
+		 * Failed to find JRE in SageTV directory, load jawt.dll by using the Windows Registry to locate the current version to use.
 		 */
 		HKEY rootKey = HKEY_LOCAL_MACHINE;
 		char currVer[16];
-		HKEY myKey;
-		DWORD readType;
+		HKEY myKey;  // myKey to NULL for RegQueryValueEx
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -272,6 +277,7 @@ jboolean JNICALL Java_sage_Sage_removeRegistryValue(JNIEnv *env,
 	const char* keyString = env->GetStringUTFChars(key, 0);
 	const char* valueNameString = env->GetStringUTFChars(valueName, 0);
 	HKEY rootKey;
+	rootKey = NULL;  // initialize to resolve C6001
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -237,6 +241,7 @@ jboolean JNICALL Java_sage_Sage_writeDwordValue(JNIEnv *env,
...
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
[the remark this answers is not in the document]
Narflex: Read up on that here: https://docs.microsoft.com/en-us/windows/win32/sync/using-mutex-objects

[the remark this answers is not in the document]
JREkiwi: But doesn't it get initial ownership via CreateMutex(NULL, TRUE, "Global\\SageTVMessagingSink") ?
https://docs.microsoft.com/en-nz/windows/win32/api/synchapi/nf-synchapi-createmutexa?redirectedfrom=MSDN

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -1431,6 +1443,7 @@ JNIEXPORT jboolean JNICALL Java_sage_Sage_releaseSystemHooks0
 	 	if ( WAIT_OBJECT_0 != dwWaitResult )
 		{
 			slog((env, "Removing system shell hook failed (object is locked)\r\n"));
+			ReleaseMutex(hMutexMessaging);
```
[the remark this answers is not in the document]
JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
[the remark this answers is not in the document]
JREkiwi: Do I just change this to `CreateMutex(NULL, FALSE, "Global\SageTVMessagingSink") `? like the other CreateMutex @ 1425

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -272,6 +275,7 @@ jboolean JNICALL Java_sage_Sage_removeRegistryValue(JNIEnv *env,
 	const char* keyString = env->GetStringUTFChars(key, 0);
 	const char* valueNameString = env->GetStringUTFChars(valueName, 0);
 	HKEY rootKey;
+	rootKey = NULL;
```
JREkiwi: Missed this one

native/dll/Win32DLL/SageTVWin32DLL.cpp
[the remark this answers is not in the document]
JREkiwi: Removed

native/dll/Win32DLL/OpenGLVideoRenderer.cpp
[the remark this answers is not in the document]
JREkiwi: Change to the same as DX9 video renderer file

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -1013,7 +1021,7 @@ void loadAWTLib()
 		*goodSlash = 0;
 		goodSlash = strrchr(jvmPath, '\\');
 		if (!goodSlash) return;
-		strcpy(goodSlash + 1, "jawt.dll");
+		strcpy_s(jvmPath, sizeof(jvmPath), "jawt.dll");
```
Narflex: Look at my prior comments here....with what you have now you just end up with 'jawt.dll' for the jvmPath....nothing is being done with all the slash stuff.

JREkiwi: Done

native/dll/Win32DLL/SageTVWin32DLL.cpp
```diff
@@ -979,7 +979,11 @@ void loadAWTLib()
 		strcat_s(includedJRE, sizeof(includedJRE), "jre\\bin\\jawt.dll");
 		sageLoadedAwtLib = LoadLibrary(includedJRE);
 		if (sageLoadedAwtLib)
+		{
+			slog(("INFO:jawt.dll path = %s\r\n", includedJRE));;
```
Narflex: Can you remove the extra parenthesis and semicolon on this one and the other slog line you added below? Or just remove those log lines...either way. Then this one will be done. :)

JREkiwi: Done semicolon.
Without the extra parenthesis it complains "cannot determine which instance of overloaded function 'sysOutPrint' is intended"
Does that mean I'm missing something else somewhere?

Narflex: If it's saying that then you can leave them, it's because of the macro...I forgot that's how we did it.

JREkiwi:
Apologies for all the back and forth and thanks once again for all the time you've spent walking me through this.

native/dll/Win32DLL/SageTVWin32DLL.cpp
JREkiwi: Done

Narflex:
No problem, thanks for the cleanups. :)

PR #432, opened 2020-03-15
Clean up of VS2015 warnings - Part 4

JREkiwi:
Changes from SageService Launcher and MiniClient Launcher applied to Launcher
Resolve Compile and Static Analysis warnings
Change to MS safe/secure functions
Reinstate field in GetMapStringValue in jni-util.cpp and jni-util.h
Tidy up indentation in WindowsServiceControl.cpp
Correct comments in MiniClientLauncher.cpp and ServiceControlLaunch.cpp

### DIFF
--- a/native/common/jni-util.cpp
+++ b/native/common/jni-util.cpp
@@ -665,7 +665,7 @@ jboolean GetMapBoolValue(JNIEnv* env, jobject map, const char* optionName, jbool
 	}
 	return JNI_TRUE;
 }
-jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, char* rv, unsigned long rvlen)
+jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, char* rv, unsigned long rvlen, unsigned long* rvCount)
 {
 	jstring keyString = env->NewStringUTF(optionName);
 	static jclass mapClass = (jclass) env->NewGlobalRef(env->FindClass("java/util/Map"));
--- a/native/exe/SageLauncher/launcher.cpp
+++ b/native/exe/SageLauncher/launcher.cpp
@@ -13,8 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Eliminate silly MS compiler security warnings about using POSIX functions
-#pragma warning(disable : 4996)
+ /*
+ *       ******************************************************************************************
+ *       Microsoft Visual Studio project codepage intentionally set to MBCS
+ *       The strings passed in the JavaVMOption struct use the platform default character encoding,
+ *       so they can't be passed as 16-bit Unicode chars.
+ *
+ *       *******************************************************************************************
+ */
+
+#pragma warning(disable : 6248) //disable warnings for NULL DACL
 
 #include <stdlib.h>
 #include <jni.h>
@@ -32,8 +40,8 @@
 LRESULT CALLBACK WndProc( HWND hWnd, UINT messg,
 								WPARAM wParam, LPARAM lParam );
 
-#define JVM_MISSING if (hWnd){MessageBox(NULL, "Could not get information on current JVM.\nPlease install Java Runtime Environment 1.7 or greater","Java Missing", MB_OK);}\
-else{OutputDebugStringA("Could not get information on current JVM.\nPlease install Java Runtime Environment 1.7 or greater");}
+#define JVM_MISSING if (hWnd){MessageBox(NULL, "Could not get information on current JVM.\nPlease install Java Runtime Environment 1.7 or higher (Java 1.8 preferred)","Java Missing", MB_OK);}\
+else{OutputDebugString("Could not get information on current JVM.\nPlease install Java Runtime Environment 1.7 or higher (Java 1.8 preferred)");}
 
 static JNIEnv* globalenv = 0;
 static HANDLE stdOutHandle = 0;
@@ -71,7 +79,7 @@ void sysOutPrint(const char* cstr, ...)
     va_list args;
     va_start(args, cstr);
     char buf[1024];
-    vsprintf(buf, cstr, args);
+    vsprintf_s(buf, sizeof(buf), cstr, args);
     va_end(args);
 	jstring jstr = env->NewStringUTF(buf);
 	static jclass cls = (jclass) env->NewGlobalRef(env->FindClass("java/lang/System"));
@@ -95,7 +103,7 @@ void sysOutPrint(JNIEnv* env, const char* cstr, ...)
     va_list args;
     va_start(args, cstr);
     char buf[1024];
-    vsprintf(buf, cstr, args);
+    vsprintf_s(buf, sizeof(buf), cstr, args);
     va_end(args);
 	jstring jstr = env->NewStringUTF(buf);
 	static jclass cls = (jclass) env->NewGlobalRef(env->FindClass("java/lang/System"));
@@ -115,8 +123,8 @@ VOID SvcDebugOut(LPSTR String, DWORD Status)
     CHAR  Buffer[1024]; 
     if (strlen(String) < 1000) 
     { 
-        sprintf(Buffer, String, Status); 
-        OutputDebugStringA(Buffer); 
+        sprintf_s(Buffer, sizeof(Buffer), String, Status);
+        OutputDebugString(Buffer); 
     } 
 }
 
@@ -128,7 +136,7 @@ void errorMsg(char* msg, char* title, HWND hWnd)
 	}
 	else
 	{
-		OutputDebugStringA(msg);
+		OutputDebugString(msg);
 	}
 }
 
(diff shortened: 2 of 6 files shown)


native/common/jni-util.cpp
```diff
@@ -665,7 +665,7 @@ jboolean GetMapBoolValue(JNIEnv* env, jobject map, const char* optionName, jbool
 	}
 	return JNI_TRUE;
 }
-jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, char* rv, unsigned long rvlen)
+jboolean GetMapStringValue(JNIEnv* env, jobject map, const char* optionName, char* rv, unsigned long rvlen, unsigned long* rvCount)
```
Narflex: I thought this was supposed to get changed in native/dll/DShowPlayer-2/PlaybackGraph.cpp#L278 in that last set of changes you did?  Did that somehow get dropped?  rvCount is unused in the only place that calls this function so we don't need to have it.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -178,20 +186,24 @@ void popupExceptionError(JNIEnv* env, jthrowable thrower, char* errTitle, HWND h
...
 
 jint JNICALL my_vfprintf(FILE *fp, const char *format, va_list args)
 {
 	if (stdOutHandle)
 	{
-	    char buf[1024];
+	    char buf[1024] = {0};
 		DWORD numWrit;
-		DWORD len = _vsnprintf(buf, sizeof(buf), format, args);
+		DWORD len = _vsnprintf_s(buf, sizeof(buf), (sizeof(buf) - strlen(buf)), format, args);
```
Narflex: The strlen(buf) call is unnecessary, the array is initialized two lines up to be empty, so this will always be zero.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -368,15 +383,17 @@ void doStageCleaning(BOOL bService, BOOL bClient)
 	}
 
 	// Handle the deletions files first
-	FILE* fp = fopen("stageddeletes.txt", "r");
+	FILE* fp;
+	fopen_s(&fp, "stageddeletes.txt", "r");
```
Narflex: If you want to switch to fopen_s (not sure why you're changing that, guessing another MS warning) then you need to check the return value from the call. That's how it indicates failure, not by setting fp to NULL. Same for all the places you switched this below.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -439,10 +456,10 @@ void doStageCleaning(BOOL bService, BOOL bClient)
 
 #define VER_SUITE_WH_SERVER 0x00008000
 
-int WINAPI WinMain( HINSTANCE hInst, 	/*Win32 entry-point routine */
-					HINSTANCE hPreInst, 
-					LPSTR lpszCmdLine, 
-					int nCmdShow )
+int WINAPI WinMain(__in HINSTANCE hInst, 	/*Win32 entry-point routine */
```
Narflex: We're not really using annotations like this anywhere that I recall, so let's not start adding them.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -637,8 +654,12 @@ int WINAPI WinMain( HINSTANCE hInst, 	/*Win32 entry-point routine */
 			}
 		}
 	}
-	chdir(folderPath);
-	delete [] folderPath;
+	errno_t err;
+	if ((err = _chdir(folderPath)) != 0) // If this doesn't return 0 we have bigger problems
+	{
+		delete[] &folderPath;
```
Narflex: This should just be 'delete [] folderPath'; Get rid of the ampersand.

Narflex: Better yet..change folderPath to be non-heap allocated (i.e. char folderPath[2048]; and use sizeof(folderPath) in place of 2048 in the GetModuleFilename call).

native/exe/SageLauncher/launcher.cpp
```diff
@@ -875,8 +895,8 @@ int WINAPI WinMain( HINSTANCE hInst, 	/*Win32 entry-point routine */
 				break;
 			}
 		} while (shouldRestartJVM(FALSE, bClient));
-		delete [] appPath;
-		delete [] newCmdLine;
+		delete [] &appPath;
```
Narflex: Remove both of these lines, you're trying to deallocate stack memory like it was heap memory. Stack memory doesn't need to be explicitly deallocated.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -993,63 +1014,70 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
...
 	{
-		errorMsg("Could not execute jvm.dll.\nPlease install Java Runtime Environment 1.7 or greater", "Java Missing", hWnd);
+		errorMsg("Could not execute jvm.dll.\nPlease install Java Runtime Environment 1.7 or higher (Java 1.8 preferred)", "Java Missing", hWnd);
 		return FALSE;
 	}
 
-	LPTSTR prefFile = new TCHAR[512];
-	LPTSTR jarPath = new TCHAR[4096];
-	LPTSTR libraryPath = new TCHAR[512];
+	size_t sizeprefFile = 512;
```
Narflex: Change all of these to stack allocations (i.e. char prefFile[512];)

native/exe/SageLauncher/launcher.cpp
```diff
@@ -993,63 +1014,70 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
...
-  strcat(jarPath, appPath);
-  strcat(jarPath, "JARs\\lucene-core-3.6.0.jar;");
-	strcat(jarPath, appPath);
-	strcat(jarPath, "xerces.jar;");
-	strcat(jarPath, appPath);
-	strcat(jarPath, "plugin.jar;");
-	strcat(jarPath, appPath);
-	strcat(jarPath, ";");
-	char* envcp = getenv("CLASSPATH");
+	strcpy_s(prefFile, sizeprefFile, bClient ? "sagetvclient " : "sagetv ");
```
Narflex: With the stack allocations above, then you can just use sizeof in all these calls to make it cleaner.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -993,63 +1014,70 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
...
+	strcat_s(jarPath, sizejarPath, appPath);
+	strcat_s(jarPath, sizejarPath, "plugin.jar;");
+	strcat_s(jarPath, sizejarPath, appPath);
+	strcat_s(jarPath, sizejarPath, (";"));
+	char* envcp = nullptr;
+	size_t sz = 0;
+	_dupenv_s(&envcp, &sz, "CLASSPATH");
 	if (envcp)
-		strcat(jarPath, envcp);
+		strcat_s(jarPath, sizejarPath, envcp);
```
Narflex: You need to add a call to free(envcp) as the documentation for _dupenv_s indicates.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -1073,34 +1101,35 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
...
 	options[vm_args.nOptions++].optionString = libraryPath;  /* set native library path */
 #endif
 #endif
 
 	// Now append everything in the JARs subfolder (that's a .jar file) to the classpath
-	char* dirPath = new char[1024];
-	strcpy(dirPath, appPath);
-	strcat(dirPath, "JARs\\*.jar");
+	size_t sizedirPath = _MAX_PATH;
+	char* dirPath = new char[sizedirPath];
```
Narflex: Again, use static rather than dynamic allocation of fixed size arrays.

native/exe/SageLauncher/launcher.cpp
```diff
@@ -1135,7 +1164,7 @@ int launchJVMSage(LPSTR lpszCmdLine, HWND hWnd, BOOL bClient, BOOL bService)
 		if (RegQueryValueEx(myKey, "JVMMaxHeapSizeMB", 0, &readType, (LPBYTE) &regMemVal, &hsize) == ERROR_SUCCESS)
 		{
 			if (regMemVal > 0)
-				sprintf(memString, "-Xmx%dm", regMemVal);
+				sprintf_s(memString, sizeof(memString), "-Xmx%dm", (int)regMemVal);
```
Narflex: You shouldn't have to cast regMemVal...instead use the proper format type in sprintf instead if it's complaining about that (I think it's %lu that it wants, or maybe %u).

JREkiwi:
Sorry, real life has intervened and I no longer have time to work on this

PR #434, opened 2020-04-17
Added the ability to create a Format parser plugin

jvl711:
Added the ability to use a Plugin for format detection of media files instead of built in ffmpeg.
Added 2160p as a Pretty resolution to VideoFormat
Added HEVC as a supported media format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 * Set default 1G heap for 64-bit (Windows)
 * Updates for OPTUS D1 transponder changes to DVB-S2
 * Fix: Schedules Direct EPG grabber failed to finish updating some satellite-based lineups
+* Added the ability to use a Plugin for format detection of media files instead of built in ffmpeg.
+* Added 2160p as a Pretty resolution to VideoFormat
+* Added HEVC as a supported media format
 
 ## Version 9.2.1 (2019-03-23)
 * 64-bit AVI playback and music fixes (Windows)
--- a/java/sage/media/format/FormatParser.java
+++ b/java/sage/media/format/FormatParser.java
@@ -231,10 +231,6 @@ public class FormatParser
         }
 
       }
-      else
-      {
-        if (sage.Sage.DBG) System.out.println("No format detector configured, or error initializing");
-      }
       
       if (sage.Sage.DBG) System.out.println("Now using external format detector for: " + f);
       String ffmpegInfo = getFFMPEGFormatInfo(f.toString());
--- a/java/sage/media/format/FormatParser.java
+++ b/java/sage/media/format/FormatParser.java
@@ -60,6 +60,7 @@ public class FormatParser
     { "MATROSKA,WEBM", MediaFormat.MATROSKA },
   };
 
+  private static FormatParserPlugin formatParserPluginInstance;
   private static final long MPEG_PARSER_SEARCH_LENGTH = 30*1024*1024;
   private static boolean DISABLE_FORMAT_DETECTION = false;
   private static boolean MINIMIZE_EXIF_MEM_USAGE = false;
@@ -192,44 +193,44 @@ public class FormatParser
             
       if(plugin != null)
       {
-        if (sage.Sage.DBG) System.out.println("Format detector configured");
+        if (sage.Sage.DBG) System.out.println("Using the format detector plugin");
         
         try
         {
-            plugin.initialize(f);
-            String formatName = plugin.getFormatName();
-            long duration = plugin.getDuration();
-            long bitrate = plugin.getBitrate();
-            BitstreamFormat streams [] = plugin.getStreamFormats();
+          plugin.initialize(f);
+          String formatName = plugin.getFormatName();
+          long duration = plugin.getDuration();
+          long bitrate = plugin.getBitrate();
+          BitstreamFormat streams [] = plugin.getStreamFormats();
 
-            if(streams != null && streams.length > 0)
-            {
-                format.setFormatName(formatName);
-                format.setDuration(duration);
-                format.setBitrate((int)bitrate);
-                format.setStreamFormats(streams);
+          if(streams != null && streams.length > 0)
+          {
+            format.setFormatName(formatName);
+            format.setDuration(duration);
+            format.setBitrate((int)bitrate);
+            format.setStreamFormats(streams);
+            addAdditionalMetadata(f, format);
 
-                return format;
-            }
+            return format;
+          }
 
         }
         catch(Exception ex)
         {
-            System.out.println("Error in Format Detector Plugin: " + ex.getMessage());
-            ex.printStackTrace();
+          System.out.println("Error in Format Detector Plugin: " + ex.getMessage());
+          ex.printStackTrace();
         }
         finally
         {
-            try
-            {
-                plugin.deconstruct();
-            }
-            catch(Exception ex)
-            {
-                System.out.println("Error deconstructing plugin: " + ex.getMessage());
-            }
+          try
+          {
+            plugin.deconstruct();
+          }
+          catch(Exception ex)
+          {
+            System.out.println("Error deconstructing plugin: " + ex.getMessage());
+          }
         }
-
       }
       
       if (sage.Sage.DBG) System.out.println("Now using external format detector for: " + f);
@@ -442,22 +443,34 @@ public class FormatParser
 
   private static FormatParserPlugin getFormatParserPluginInstance()
   {
-    FormatParserPlugin parsie = null;
-    String parsePlugin = sage.Sage.get("mediafile_mediaformat_parser_plugin", "");
-
-    try
+    /*
+     * Check to see if an instance has already been created.  If not check to see if one is configured and attempt to create an instance
+    */
+    if(FormatParser.formatParserPluginInstance == null)
     {
-        parsie = (FormatParserPlugin) Class.forName(parsePlugin, true, sage.Sage.extClassLoader).newInstance();
-    } 
-    catch (Throwable e1)
-    {
-        if (sage.Sage.DBG)
+      String parsePlugin = sage.Sage.get("mediafile_mediaformat_parser_plugin", "");
+      
+      if(parsePlugin.isEmpty())
+      {
+        FormatParser.formatParserPluginInstance = null;
+      }
+      else
+      {
+        try
         {
-            System.out.println("Error instantiating metadata parser plugin of:" + e1);
+          FormatParser.formatParserPluginInstance = (FormatParserPlugin) Class.forName(parsePlugin, true, sage.Sage.extClassLoader).newInstance();
+        } 
+        catch (Throwable e1)
+        {
+          if (sage.Sage.DBG)
+          {
+              System.out.println("Error instantiating metadata parser plugin of: " + e1);
+          }
         }
+      }
     }
 
-    return parsie;
+    return FormatParser.formatParserPluginInstance;
   }
   
   public static java.util.Map extractMetadataProperties(java.io.File mediaPath, java.io.File propFile, boolean invokePlugins)
--- a/java/sage/media/format/FormatParser.java
+++ b/java/sage/media/format/FormatParser.java
@@ -15,8 +15,6 @@
  */
 package sage.media.format;
 
-import java.util.concurrent.locks.ReentrantLock;
-
 /**
  *
  * @author Narflex
@@ -63,7 +61,7 @@ public class FormatParser
   };
 
   private static FormatParserPlugin formatParserPluginInstance;
-  private static Object  formatParserPluginLock = new Object();
+  private static Object formatParserPluginLock = new Object();
   private static final long MPEG_PARSER_SEARCH_LENGTH = 30*1024*1024;
   private static boolean DISABLE_FORMAT_DETECTION = false;
   private static boolean MINIMIZE_EXIF_MEM_USAGE = false;
@@ -192,7 +190,7 @@ public class FormatParser
         }
       }
 
-      FormatParserPlugin plugin = FormatParser.getFormatParserPluginInstance();
+      FormatParserPlugin plugin = getFormatParserPluginInstance();
             
       if (plugin != null)
       {
@@ -428,26 +426,22 @@ public class FormatParser
     /*
      * Check to see if an instance has already been created.  If not check to see if one is configured and attempt to create an instance
     */
-    if (FormatParser.formatParserPluginInstance == null)
+    if (formatParserPluginInstance == null)
     {
       String parsePlugin = sage.Sage.get("mediafile_mediaformat_parser_plugin", "");
 
-      if (parsePlugin.isEmpty())
-      {
-        FormatParser.formatParserPluginInstance = null;
-      }
-      else
+      if (!parsePlugin.isEmpty())
       {
         synchronized (formatParserPluginLock)
         {
-          if (FormatParser.formatParserPluginInstance != null)
+          if (formatParserPluginInstance != null)
           {
-              return FormatParser.formatParserPluginInstance;
+              return formatParserPluginInstance;
           }
           
           try
           {
-            FormatParser.formatParserPluginInstance = (FormatParserPlugin) Class.forName(parsePlugin, true, sage.Sage.extClassLoader).newInstance();
+            formatParserPluginInstance = (FormatParserPlugin) Class.forName(parsePlugin, true, sage.Sage.extClassLoader).newInstance();
           } 
           catch (Throwable e1)
           {
@@ -460,7 +454,7 @@ public class FormatParser
       }
     }
     
-    return FormatParser.formatParserPluginInstance;
+    return formatParserPluginInstance;
   }
   
   public static java.util.Map extractMetadataProperties(java.io.File mediaPath, java.io.File propFile, boolean invokePlugins)
--- a/java/sage/media/format/FormatParser.java
+++ b/java/sage/media/format/FormatParser.java
@@ -191,21 +191,21 @@ public class FormatParser
 
       FormatParserPlugin plugin = FormatParser.getFormatParserPluginInstance();
             
-      if(plugin != null)
+      if (plugin != null)
       {
         if (sage.Sage.DBG) System.out.println("Using the format detector plugin");
         
         try
         {
-          ContainerFormat pluingFormat = plugin.parseFormat(f);
+          ContainerFormat pluginFormat = plugin.parseFormat(f);
             
-          if(pluingFormat.streamFormats != null && pluingFormat.streamFormats.length > 0)
+          if (pluginFormat.streamFormats != null && pluginFormat.streamFormats.length > 0)
           {
-            addAdditionalMetadata(f, pluingFormat);
-            return pluingFormat;
+            addAdditionalMetadata(f, pluginFormat);
+            return pluginFormat;
           }
         }
-        catch(Exception ex)
+        catch (Exception ex)
         {
           System.out.println("Error in Format Detector Plugin: " + ex.getMessage());
           ex.printStackTrace();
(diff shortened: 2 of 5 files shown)


Narflex:
What's the motivation for this change?  You're making a plugin interface for something that should be consistent and standard...if there's a problem with format detection, then it should be fixed in the existing format detector itself.

jvl711:
@Narflex 

I wanted to remove the reliance on the core to be able to identify newer formats.  I also wanted to make a method that was a little faster and more efficient.

I created JNI wrapper around ffmpeg, which gives a lot more access to additional properties.  It is very fast, and detects a lot of additional formats that the current version of ffmpeg in the sage core does not.  I believe that updating the JNI version may be much quicker.

SageTranscoder does not know about HEVC/H265.  There are a number of subtitle formats it does not detect.  It does not display if a subtitle track is forced.  Certain files take a considerable amount of time for SageTranscoder to identify.

The JNI version I wrote could be folded into sage, but honestly I am not sure how to make the c portion compile for both windows and linux.  I could send you the code if you were interested in directing me on how to get it properly implemented into sage.  I would be willing to give it a try.

Also updating the SageTranscoder.exe seems to be a pretty daunting task.  It handles a lot more than just format detection, and is a really old version of ffmpeg.  A handful of people have stated they were going to move it to a newer library and never did.

My feeling was that if it was not completely tied to the core it could be upgraded more quickly/easily/independently.

Narflex:
One of the main reasons we run FFMPEG as another process and not through JNI directly (except for when we're using it as a demux in DirectShow) is that there were way too many times it would just hang up forever processing files...so we needed a clean way to terminate it. Also, when it would crash we didn't want it to bring down SageTV as well.  So I wouldn't want to convert SageTV over to that type of design where FFMPEG was used in-process for format detection.

I do agree that allowing format detection to be more up to date, without having to update the old version of FFMPEG would be very useful.

What's your plan for the format detector plugin you've written? Are you going to make a GitHub project with that in there, initially only supporting Windows, and then try to get help from the community to extend it to Linux support? (I don't have the time to do that unfortunately)  I want to be sure changes to the SageTV project benefit users in general and aren't just for one person.

jvl711:
I agree with your concern about the JNI.  When I started working with it I was not aware that it was not isolated from the main java process.  I caused Sage to crash on a number of occasions while developing the plugin.  I do think that the ffmpeg library is much more matrue now and I do not think it has a lot of those issues anymore.

I have been using this code for a few months without any issues "fingers crossed".  It seems pretty stable, even with random downloaded content.

I would like to post the plugin and ask for some assistance on the linux part.  I do not think it would be much effort.  If this is not a direction you want to go, I completely understand.

I just have a lot of code that I would like to share.  It just takes a lot more time to share than to just write a bunch of code and enjoy it yourself.... :/

Narflex:
I would want to keep it out of process though for stock SageTV (but if you had a plugin where people voluntarily could run it in process, I won't object to that).  One thing I don't want is any default code going into SageTV which could impact stability...it's pretty damn stable the way it is and I'm quite proud of that. :)  I do want to encourage you to post the plugin to get help on Linux...I'm just saying I won't be doing all that work...but I'd definitely take a look and offer advice.

And yeah, it does take time to share open source code...but that's the deal I'll make...you can go ahead and put this interface into SageTV as long as you agree to put up the code for you implementation as open source. :)

CraziFuzzy:
Been away from Sage Development (or any development for that matter), but saw this notificaiton come up in my email and thought I'd check it out.  I think i was one that was wanting to update SageTVTranscoder at some time in the past, and wished back then that it could be interfaced with a 'stock' and updated version of ffmpeg.exe at some point.  Is it possible for this wrapper to still be run outside the main SageTV process?  That would allow the modularity of it being a plugable upgrade, without it being stuck within the core java process.

jvl711:
@CraziFuzzy 

Yes.  This would allow you to use any method you want to do Format Detection.  If you look at the plugin Interface you will see it is fairly straight forward.  It has methods for higher level info about the file, and then it has a method to pass info about all of the streams in the file.  You can use any method you want to populate.

The method I am currently using is a JNI wrapper around a more recent version of FFmpeg.  I took this approach for a number of reasons.

1. Wanted to learn JNI
2. Wanted to learn FFmpeg library a little better
3. Wanted to have access to FFmpeg inside of java for my personal future development

But as Narflex has stated, it may be better to isolate this processing to an exe.  Sage currently does stdout scrapping of FFmpeg.  Someone could replicate that with a newer version with this plugin if they wanted.

Josh

jvl711:
@Narflex 
 I agree to publish my code that utilizes this plugin to github.  I started working on converting the svn to git and cleaning the code for submission to github this morning.

Thank you for working with me!

Josh

FYI...  I have been a continuous Sage user for over 16 years.  I have been lurking around, and occasionally sharing my work with the community.  Modifying and playing with Sage is probably one of my favorite hobbies.  Thank you for all you have done!

jvl711:
@Narflex 

Published the first portion of the code.  https://github.com/jvl711/JavaFFmpegLibrary

I will publish the library that has the plugin in it next.  It is contained in a much larger library, so it might take a little additional time to put together.

Josh

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sage.media.format;
+
+
```
Narflex, line 18: Please add a description of how the plugin should be used. I'd also suggest changing the overall design for efficiency. It shouldn't have to instantiate a new class using the classloader for every file it wants to parse. I'd suggest this be a one method interface with parseFormat(File) and then add a new inner class to the plugin of something like ParsedFormat which just the 4 fields that would otherwise be returned from the method calls. Then FormatParser can hold onto a copy of it statically.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 * Set default 1G heap for 64-bit (Windows)
 * Updates for OPTUS D1 transponder changes to DVB-S2
 * Fix: Schedules Direct EPG grabber failed to finish updating some satellite-based lineups
+* Added the ability to use a Plugin for format detection of media files instead of built in ffmpeg.
+* Added 2160p as a Pretty resolution to VideoFormat
+* Added HEVC as a supported media format
```
Narflex, line 15: You didn't include the new HEVC constant.

jvl711: Good Catch.  I committed that change to the Branch/Pull request.  It is in MediaFormat class

java/sage/media/format/FormatParserPlugin.java
jvl711: What if I only create one static instance of the plugin in FormatParser, but leave the FormatParserPlugin interface the way that it is?

The initialize and deconstruct methods was meant to handle only having one instance of the plugin.  I think it will work the way you want.  Was there another reason to only have one parseFormat method.

Narflex: It's not thread safe if you do it that way where you're reusing the object like that w/ the existing interface....if you made it just one method (and it didn't maintain internal state, which there's really no reason to across multiple invocations) then that problem would go away.  Having a pattern with init/destroy like this for something that is not intended to maintain any long term state doesn't really seem necessary either.

jvl711: Understand.  I was trying to walk the implemetor down the path, but I get what you are saying.

Should I just expose one method in the interface that accepts a file and returns the ContainerFormat?  Than the implementation of the plug-in will handle setting the proper properties?

Or would you prefer that I make a new structure that they need to construct with all of the required data? And return that new class from the one parseFormat method.  I am just concerned the developer was going to implement it correctly.

Let me know what you think.

Narflex: Yeah, you can just have it return ContainerFormat directly....I don't foresee that changing so it's fine to consider it a stable interface. So just:

ContainerFormat parseFormat(File f);

jvl711:
@Narflex 

I think I addresses all of your comments let me know what you think.

Thanks,
Josh

java/sage/media/format/FormatParser.java
```diff
@@ -187,6 +189,29 @@ public static ContainerFormat getFileFormat(java.io.File f)
         }
       }
 
+      FormatParserPlugin plugin = FormatParser.getFormatParserPluginInstance();
+            
+      if(plugin != null)
```
Narflex, line 194: Put a space after if and before the paren

java/sage/media/format/FormatParser.java
```diff
@@ -187,6 +189,29 @@ public static ContainerFormat getFileFormat(java.io.File f)
...
+      {
+        if (sage.Sage.DBG) System.out.println("Using the format detector plugin");
+        
+        try
+        {
+          ContainerFormat pluingFormat = plugin.parseFormat(f);
+            
+          if(pluingFormat.streamFormats != null && pluingFormat.streamFormats.length > 0)
```
Narflex, line 202: Space after if and also check that pluginFormat is not null.

java/sage/media/format/FormatParser.java
```diff
@@ -187,6 +189,29 @@ public static ContainerFormat getFileFormat(java.io.File f)
...
+          {
+            addAdditionalMetadata(f, pluingFormat);
+            return pluingFormat;
+          }
+        }
+        catch(Exception ex)
```
Narflex, line 208: Space after catch, and catch a Throwable instead of Exception so nothing gets through.

java/sage/media/format/FormatParser.java
```diff
@@ -187,6 +189,29 @@ public static ContainerFormat getFileFormat(java.io.File f)
...
 
+      FormatParserPlugin plugin = FormatParser.getFormatParserPluginInstance();
+            
+      if(plugin != null)
+      {
+        if (sage.Sage.DBG) System.out.println("Using the format detector plugin");
+        
+        try
+        {
+          ContainerFormat pluingFormat = plugin.parseFormat(f);
```
Narflex, line 200: Misspelled pluginFormat

java/sage/media/format/FormatParser.java
```diff
@@ -395,6 +420,38 @@ private static boolean addAdditionalMetadata(java.io.File f, ContainerFormat for
...
+    {
+      String parsePlugin = sage.Sage.get("mediafile_mediaformat_parser_plugin", "");
+      
+      if(parsePlugin.isEmpty())
+      {
+        FormatParser.formatParserPluginInstance = null;
+      }
+      else
+      {
+        try
```
Narflex, line 438: This isn't thread-safe, so either put a static lock in here...or put a comment that explains we know that it's not thread-safe and it's OK for one plugin instance to clobber another if there's a race condition.

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,75 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sage.media.format;
+
+/**
+ * This class provides an interface for the creation of a plug to override
+ * the built in file format detection.  This will allow the community to create
+ * plugins that will allow SageTV to be able to recognize newer media formates
+ * that the core is not currently able to handle.
```
Narflex, line 22: You might also want to add that it will not override format detection for MP3, image or MPEG2-TS/PS files that the internal format detector can handle.  I'd like to keep it that way, so don't just move where it's invoked in the FormatParser instead. :)  It's really designed to replace the existing FFMPEG format detector.

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,75 @@
...
+ * that the core is not currently able to handle. 
+ * 
+ * @author jvl711
+ */
+public interface FormatParserPlugin 
+{
+
+  /**
+   * SageTV will pass a file to this method when it needs to know its format.  This 
+   * could because it is the first time it saw the file, or because SageTV needed to
```
Narflex, line 31: "could be because"

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,75 @@
...
+   * <code>
+   * ContainerFormat.setStreamFormats(streams);
+   * </code>
+   * 
+   * </Pre>
+   * If you return a container format with no streams, SageTV will discard your ContainerFormat and use it's own format
+   * detector to attempt to determine what the files format is.
+   * 
+   * @param file Media file to determine the media format for
+   * @return ContainerFormat populated with the format of the given file
```
Narflex, line 71: Should also mention it can return null or a ContainerFormat with no streams in case it is unable to detect the format.

java/sage/media/format/FormatParser.java
jvl711: Would be OK if I made the method synchronized?  I would think this would not get called by more than one thread very often.

java/sage/media/format/FormatParserPlugin.java
jvl711: I agree with this completely, and that was my intent.  It looked like a lot of effort was put into handling MPEG2.  I did only want to override the ffmpeg portion.  I will update the comment to try and make that more clear.

jvl711:
@Narflex 

I think I addressed all of the comments.  Let me know how it looks.

Thanks,
Josh

Narflex:
Please look through my comments again....a few of them were not addressed. And rather than making that method static synchronized, you should create a static Object which you synchronize on in there only if the plugin doesn't exist and has a property and you are about to create it (and then once in the block, if you see the plugin exists just return with it rather than creating a duplicate). That'll minimize locking as much as possible.

jvl711:
@Narflex 

I am a little confused on how you want to protect against multiple formatParserPluginInstance.

Do you want me to use a lock object instead of synchronized to protect the critical area?  I think you need to lock around the whole if block that checks for null if you want to make sure that we do not instantiate formatParserPlugin twice.

Could you give me a little more insight,

Josh

java/sage/media/format/FormatParser.java
```diff
@@ -57,13 +57,15 @@
     { "dtsc / 0x63737464", MediaFormat.DTS },
     { "mlpa / 0x61706c6d", MediaFormat.DOLBY_HD },
     { "AC-3 / 0X332D6361", MediaFormat.AC3 },
+    { "MATROSKA,WEBM", MediaFormat.MATROSKA },
   };
 
+  private static FormatParserPlugin formatParserPluginInstance;
```
Narflex, line 63: For the lock, add another line such as:

private static Object formatParserPluginLock = new Object();

java/sage/media/format/FormatParser.java
```diff
@@ -395,6 +420,38 @@ private static boolean addAdditionalMetadata(java.io.File f, ContainerFormat for
     return rv;
   }
 
+  private synchronized static FormatParserPlugin getFormatParserPluginInstance()
```
Narflex, line 423: Update this method to be:

```
private static FormatParserPlugin getFormatParserPluginInstance()
{
    /*
     * Check to see if an instance has already been created.  If not check to see if one is configured and attempt to create an instance
    */
    if (formatParserPluginInstance == null)
    {
      String parsePlugin = sage.Sage.get("mediafile_mediaformat_parser_plugin", "");

      if (parsePlugin.isEmpty())
      {
          return null;
      }
      else
      {
    synchronized (formatParserPluginLock)
    {
      if (formatParserPluginInstance != null)
        return formatParserPluginInstance;
        try        
        {
         formatParserPluginInstance = (FormatParserPlugin) Class.forName(parsePlugin, true, sage.Sage.extClassLoader).newInstance();
        } 
        catch (Throwable e1)
        {
          if (sage.Sage.DBG)
          {
              System.out.println("Error instantiating metadata parser plugin of: " + e1);
          }
        }
      }
    }
  }

    return formatParserPluginInstance;
  }

```

jvl711: @Narflex  never mind.  I see the second check for null....  Sorry

jvl711:
@Narflex 

OK.  I think I have everything addresses.  Let me know if I have to go over it again.  Sorry I missed a handful of those items.

I am assuming you wanted the lock code designed that way so a majority of the time it will not enter the lock area.  That does make a lot of sense after seeing it.  I was not sure what you were going for with that.  I kept getting stuck on the checking for null not being in the lock region.  Was not thinking to just add the check for null again in the lock area. 

Josh

java/sage/media/format/FormatParser.java
```diff
@@ -15,6 +15,8 @@
  */
 package sage.media.format;
 
+import java.util.concurrent.locks.ReentrantLock;
```
Narflex, line 18: Remove this since you aren't using this class.

java/sage/media/format/FormatParser.java
```diff
@@ -187,6 +192,29 @@ public static ContainerFormat getFileFormat(java.io.File f)
         }
       }
 
+      FormatParserPlugin plugin = FormatParser.getFormatParserPluginInstance();
```
Narflex, line 195: Having the "FormatParser." prefix isn't needed, we are in that class already.

java/sage/media/format/FormatParser.java
```diff
@@ -395,6 +423,46 @@ private static boolean addAdditionalMetadata(java.io.File f, ContainerFormat for
     return rv;
   }
 
+  private static FormatParserPlugin getFormatParserPluginInstance()
+  {
+    /*
+     * Check to see if an instance has already been created.  If not check to see if one is configured and attempt to create an instance
+    */
+    if (FormatParser.formatParserPluginInstance == null)
```
Narflex, line 431: Don't need FormatParser. prefix

java/sage/media/format/FormatParser.java
```diff
@@ -395,6 +423,46 @@ private static boolean addAdditionalMetadata(java.io.File f, ContainerFormat for
...
+    {
+      String parsePlugin = sage.Sage.get("mediafile_mediaformat_parser_plugin", "");
+
+      if (parsePlugin.isEmpty())
```
Narflex, line 435: Invert this to be "if (!parsePlugin.isEmpty())" and jump right into the else block (i.e. delete the next 4 lines of code).

java/sage/media/format/FormatParser.java
```diff
@@ -395,6 +423,46 @@ private static boolean addAdditionalMetadata(java.io.File f, ContainerFormat for
...
+      {
+        FormatParser.formatParserPluginInstance = null;
+      }
+      else
+      {
+        synchronized (formatParserPluginLock)
+        {
+          if (FormatParser.formatParserPluginInstance != null)
```
Narflex, line 443: Remove the FormatParser. prefix here and the next few lines below and in the return statement.

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sage.media.format;
+
+/**
+ * This class provides an interface for the creation of a plug to override
```
Narflex, line 19: "for the creation of a plugin" (not plug)

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+ * the built in file format detection.  This will allow the community to create
+ * plugins that will allow SageTV to be able to recognize newer media formates
```
Narflex, line 21: s/formates/formats

java/sage/media/format/FormatParserPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+ * that the core is not currently able to handle.
+ * 
+ * This plugin interaface does not allow for overriding the built in formate detection
```
Narflex, line 24: s/interaface/interface
s/formate/format

java/sage/media/format/FormatParser.java
```diff
@@ -57,13 +59,16 @@
     { "dtsc / 0x63737464", MediaFormat.DTS },
     { "mlpa / 0x61706c6d", MediaFormat.DOLBY_HD },
     { "AC-3 / 0X332D6361", MediaFormat.AC3 },
+    { "MATROSKA,WEBM", MediaFormat.MATROSKA },
   };
 
+  private static FormatParserPlugin formatParserPluginInstance;
+  private static Object  formatParserPluginLock = new Object();
```
Narflex, line 66: Remove the extra space after Object

jvl711:
@Narflex 

I think I addressed all of the comments.  Let me know how it looks.

Josh

Narflex:
Thanks!

PR #435, opened 2020-04-18
Add forced subtitle property to media file

jvl711:
Added forced property to the subtitle tracks.  Added code to auto select forced subtitle tracks that match the default audio language.  Added the forced subtitle property to output in the API.

Also added netbeans project files to the .gitignore files.

### DIFF
--- a/java/sage/MediaFile.java
+++ b/java/sage/MediaFile.java
@@ -1513,6 +1513,8 @@ public class MediaFile extends DBObject implements SegmentedFile
             return Integer.toString(sf[streamIdx].getOrderIndex());
           else if (lcName.equals("id"))
             return sf[streamIdx].getId();
+          else if (lcName.equals("forced"))
+            return sf[streamIdx].getForced() ? "true" : "false";
         }
       }
       else if (lcName.equals("container"))
--- a/java/sage/MediaFile.java
+++ b/java/sage/MediaFile.java
@@ -1514,7 +1514,7 @@ public class MediaFile extends DBObject implements SegmentedFile
           else if (lcName.equals("id"))
             return sf[streamIdx].getId();
           else if (lcName.equals("forced"))
-            return sf[streamIdx].getForced() ? "true" : "false";
+            return Boolean.toString(sf[streamIdx].getForced());
         }
       }
       else if (lcName.equals("container"))
--- a/java/sage/VideoFrame.java
+++ b/java/sage/VideoFrame.java
@@ -19,7 +19,6 @@ import sage.media.sub.RollupAnimation;
 
 import java.util.ArrayList;
 import java.util.List;
-import sage.media.format.SubpictureFormat;
 
 /*
  * NOTE: I use Sage.preferredServer when specifying the hostname to get around the
--- a/java/sage/VideoFrame.java
+++ b/java/sage/VideoFrame.java
@@ -19,6 +19,7 @@ import sage.media.sub.RollupAnimation;
 
 import java.util.ArrayList;
 import java.util.List;
+import sage.media.format.SubpictureFormat;
 
 /*
  * NOTE: I use Sage.preferredServer when specifying the hostname to get around the
@@ -5650,6 +5651,35 @@ public final class VideoFrame extends BasicVideoFrame implements Runnable
 
   private boolean selectDefaultSubpicLanguage()
   {
+	String defaultAudioLang = uiMgr.get("default_audio_language", "English"); 
+    MediaLangInfo mli = (MediaLangInfo) mediaLangMap.get(defaultAudioLang);
+    SubpictureFormat [] subs = this.currFile.getFileFormat().getSubpictureFormats();
+
+    for(int i = 0; i < subs.length; i++)
+    {
+      if(subs[i].getForced())      
+      {
+        //Check for a match on the two letter lang code
+        if(subs[i].getLanguage().equalsIgnoreCase(mli.twoChar))
+        {
+          System.out.println("Setting forced subtitle to index: " + i);
+          playbackControl(DVD_CONTROL_SUBTITLE_CHANGE, i, -1);
+          return true;
+        }
+
+        //Check for a match on the three letter lang code
+        for(int j = 0; j < mli.threeChar.length; j++)
+        {
+          if(subs[i].getLanguage().equalsIgnoreCase(mli.threeChar[j]))
+          {
+            System.out.println("Setting forced subtitle to index: " + i);
+            playbackControl(DVD_CONTROL_SUBTITLE_CHANGE, i, -1);
+            return true;
+          }
+        }
+      }
+    }  
+	
     String defaultLang = uiMgr.get("default_subpic_language", "");
     if (defaultLang == null || defaultLang.length() == 0)
       return true; // preferred no subtitles
--- a/java/sage/VideoFrame.java
+++ b/java/sage/VideoFrame.java
@@ -5651,43 +5651,42 @@ public final class VideoFrame extends BasicVideoFrame implements Runnable
 
   private boolean selectDefaultSubpicLanguage()
   {
-	String defaultAudioLang = uiMgr.get("default_audio_language", "English"); 
-    MediaLangInfo mli = (MediaLangInfo) mediaLangMap.get(defaultAudioLang);
-    SubpictureFormat [] subs = this.currFile.getFileFormat().getSubpictureFormats();
-
-	/*
-	 * Loop thru all subtitle tracks of the mediafile to see if there is a forced subtitle 
-	 * track that matches the configured default audio language.  If it does it will 
-	 * select the track for playback.
-    */	 
-    for(int i = 0; i < subs.length; i++)
+    String defaultAudioLang = uiMgr.get("default_audio_language", "English"); 
+    MediaLangInfo mli = (MediaLangInfo) mediaLangMap.get(defaultAudioLang);	
+    
+    String defaultLang = uiMgr.get("default_subpic_language", "");
+    
+    //Check for forced subtitles if there is not a defaul subtitle language set
+    if(defaultLang == null || defaultLang.length() == 0)
     {
-      if(subs[i].getForced())      
-      {
-        //Check for a match on the two letter lang code
-        if(subs[i].getLanguage().equalsIgnoreCase(mli.twoChar))
-        {
-          System.out.println("Setting forced subtitle to index: " + i);
-          playbackControl(DVD_CONTROL_SUBTITLE_CHANGE, i, -1);
-          return true;
-        }
+      String [] subs = getDVDAvailableSubpictures();
+      boolean hasForcedSub = false;        
 
-        //Check for a match on the three letter lang code
-        for(int j = 0; j < mli.threeChar.length; j++)
+      /*
+      * Loop thru all subtitle tracks to see if there is a forced subtitle track. 
+      * Clear the tracks that are not forced
+      */
+      for(int i = 0; i < subs.length; i++)
+      {
+        if(subs[i].toLowerCase().indexOf("[forced]") != -1)
         {
-          if(subs[i].getLanguage().equalsIgnoreCase(mli.threeChar[j]))
-          {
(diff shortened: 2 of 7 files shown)


java/sage/MediaFile.java
```diff
@@ -1513,6 +1513,8 @@ else if (lcName.equals("index"))
             return Integer.toString(sf[streamIdx].getOrderIndex());
           else if (lcName.equals("id"))
             return sf[streamIdx].getId();
+          else if (lcName.equals("forced"))
+            return sf[streamIdx].getForced() ? "true" : "false";
```
Narflex, line 1517: Make this:

return Boolean.toString(sf[streamIdx].getForced());

java/sage/VideoFrame.java
```diff
@@ -5650,6 +5651,40 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
 
   private boolean selectDefaultSubpicLanguage()
   {
+	String defaultAudioLang = uiMgr.get("default_audio_language", "English");
```
Narflex, line 5654: Fix indentatino

java/sage/VideoFrame.java
```diff
@@ -5650,6 +5651,40 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
...
+	 * Loop thru all subtitle tracks of the mediafile to see if there is a forced subtitle 
+	 * track that matches the configured default audio language.  If it does it will 
+	 * select the track for playback.
+    */	 
+    for(int i = 0; i < subs.length; i++)
+    {
+      if(subs[i].getForced())      
+      {
+        //Check for a match on the two letter lang code
+        if(subs[i].getLanguage().equalsIgnoreCase(mli.twoChar))
```
Narflex, line 5668: You need to handle the case where mli is null.

java/sage/VideoFrame.java
```diff
@@ -5650,6 +5651,40 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
...
+	String defaultAudioLang = uiMgr.get("default_audio_language", "English"); 
+    MediaLangInfo mli = (MediaLangInfo) mediaLangMap.get(defaultAudioLang);
+    SubpictureFormat [] subs = this.currFile.getFileFormat().getSubpictureFormats();
+
+	/*
+	 * Loop thru all subtitle tracks of the mediafile to see if there is a forced subtitle 
+	 * track that matches the configured default audio language.  If it does it will 
+	 * select the track for playback.
+    */	 
+    for(int i = 0; i < subs.length; i++)
```
Narflex, line 5663: I'm thinking you should do the forced subtitle selection only if they user hasn't configured a default subtitle. Otherwise, if they always use subtitles because they're hard of hearing, then your new code will break that and they'll only see the forced subs rather than all the subs.  Also, in the default subtitle selection code, you should update that to prefer the non-forced subs for the matching language in order to improve how that code works now that we know which are which.

java/sage/VideoFrame.java
Narflex: Actually, why aren't you calling performLanguageMatch with a modified subpicture array? (i.e. same length so the indices match, but then put empty strings for the non-forced subs)

java/sage/media/format/SubpictureFormat.java
```diff
@@ -23,8 +23,8 @@
 {
   public String toString()
   {
-    return "Subpic[" + formatName + (id != null ? (" id=" + id) : "") + (path != null ? (" path=" + path) : "") +
-        (language == null ? "]" : (" " + language + "]"));
+    return "Subpic[" + formatName + (id != null ? (" id=" + id) : "") + (path != null ? (" path=" + path) : "")
+                + (language == null ? "" : (" " + language)) + " forced=" + forced + "]";
```
Narflex, line 27: Change where you have:

" forced=" + forced

to be

(forced ? " forced" : "")

java/sage/media/format/SubpictureFormat.java
```diff
@@ -76,6 +86,18 @@ public String getFullPropertyString()
       sb.append(ancillaryPageId);
       sb.append(';');
     }
+    if(this.forced == true)
```
Narflex, line 89: Just change this whole block to be:

sb.append("forced=");
sb.append(forced);
sb.append(';');

java/sage/VideoFrame.java
jvl711: `
public String[] getDVDAvailableSubpictures()
  {
    sage.media.sub.SubtitleHandler subby = subHandler;
    if (subby != null && !subby.mpControlled())
    {
      return subby.getLanguages();
    }
    MediaPlayer mp = player;
    if (mp instanceof DVDMediaPlayer)
      return ((DVDMediaPlayer) mp).getDVDAvailableSubpictures();
    else
      return Pooler.EMPTY_STRING_ARRAY;
  }
`
Should I be concerned with what this code is doing?  I was having some difficulty in trying to figure out what subtitle tracks you are pulling back.  I assume from this code that depending on what player is active you may ask the player vs using the info from the format detection.  Am I correct in that assumption?  Should I only do my logic in certain cases?

When is the SubtitleHandler used vs the DVDMediaPlayer?  Should I only do my logic if using Subtitle Handler?

Narflex: It's not really about what player is active, it's about whether the subtitle tracks are stored internally with the file itself or are external files in the same directory. However, looking at the code...I can't figure out why it was done this way because they end up calling the same thing on the ContainerFormat in the end (but I'm pretty sure there's another subtle difference I'm forgetting).  That's why I'd recommend making this as aligned with how the current code operates rather than trying to do your own logic for figuring out what the track options are.

jvl711: @Narflex 

I agree with your comments, I just want to make sure I take the right approach on getting the subtitle tracks.

Should I continue to just use this.currFile.getFileFormat().getSubpictureFormats(); to get the subtitle tracks, or do I need to reverse engineer, and possibly make modifications to SubtitleHandler and all MediaPlayers that implement DVDMediaPlayer to make new methods that return forcedSubtitle tracks?

I took the easy approach because I was having difficulty tracing all of the places that Sage is getting subtitle tracks.  And this change only really impacts Subtitles inside the media file.

Josh

Narflex: There's no reason you need to change SubtitleHandler or any of the MediaPlayer code, it's outside the scope of what you're doing here.  The approach I would take is to modify ContainerFormat.getSubpictureStreamSelectionDescriptors so that it also appends onto the end " [Forced]" for forced sub tracks.  Then in VideoFrame.selectDefaultSubpicLanguage I would make a copy of the array of subpic tracks from getDVDAvailableSubpictures(), go through that array and set to the empty string all elements that don't have the "[Forced]" string in it.  Then call performLanguageMatch(defaultAudioLang, modifiedSubpicArray, DVD_CONTROL_SUBTITLE_CHANGE, getDVDSubpicture()) and if that doesn't return true, then proceed with normal default subpic selection....you could also skip that step if you don't find any Forced subs in that list.

jvl711:
@Narflex 

I think I addressed everything except for:

> Also, in the default subtitle selection code, you should update that to prefer the non-forced subs for the matching language in order to improve how that code works now that we know which are which.

I will look at that next.  If you have a minute can you review what I have so far.  

Thanks,
Josh

java/sage/VideoFrame.java
```diff
@@ -5650,9 +5651,42 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
 
   private boolean selectDefaultSubpicLanguage()
   {
+    String defaultAudioLang = uiMgr.get("default_audio_language", "English"); 
+    MediaLangInfo mli = (MediaLangInfo) mediaLangMap.get(defaultAudioLang);
```
Narflex, line 5655: Remove this line.

java/sage/VideoFrame.java
```diff
@@ -5650,9 +5651,42 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
...
+    MediaLangInfo mli = (MediaLangInfo) mediaLangMap.get(defaultAudioLang);	
+    
     String defaultLang = uiMgr.get("default_subpic_language", "");
-    if (defaultLang == null || defaultLang.length() == 0)
-      return true; // preferred no subtitles
+    
+    //Check for forced subtitles if there is not a defaul subtitle language set
+    if(defaultLang == null || defaultLang.length() == 0)
+    {
+      String [] subs = getDVDAvailableSubpictures();
```
Narflex, line 5662: Make a copy of this array, otherwise you're overwriting a cache used in other areas potentially.

java/sage/VideoFrame.java
```diff
@@ -5650,9 +5651,42 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
...
+      boolean hasForcedSub = false;        
+
+      /*
+      * Loop thru all subtitle tracks to see if there is a forced subtitle track. 
+      * Clear the tracks that are not forced
+      */
+      for(int i = 0; i < subs.length; i++)
+      {
+        if(subs[i].toLowerCase().indexOf("[forced]") != -1)
```
Narflex, line 5671: Put a Sage.rez() around the "forced" text for localization purposes.

java/sage/VideoFrame.java
```diff
@@ -5650,9 +5651,42 @@ private int getMatchingLangIndex(String[] langOptions, String targetLang)
...
+        }
+        else
+        {
+          subs[i] = "";  
+        }
+      }
+        
+      if(hasForcedSub)
+      {
+        return performLanguageMatch(defaultAudioLang, subs, DVD_CONTROL_SUBTITLE_CHANGE, getDVDSubpicture());
```
Narflex, line 5683: Don't just return here, if this returns false then you should fall-through to the normal subpicture selection.

java/sage/media/format/ContainerFormat.java
```diff
@@ -737,7 +737,7 @@ public String getPrimaryAudioFormat()
       rv[i] = "# " + (i + 1);
       String lang = afs[i].getLanguage();
       if (lang != null && lang.length() > 0)
-        rv[i] += " " + lang;
+        rv[i] += " " + lang + (afs[i].getForced() ? " [Forced]" : "");
```
Narflex, line 740: Put a Sage.rez() around "Forced"

java/sage/media/format/SubpictureFormat.java
Narflex: You missed this comment....

java/sage/VideoFrame.java
jvl711: I have already checked above that there is no "default_subpic_language".  I believe we would be done if it was unable to find a match.  I believe we should be OK.  Am I wrong?

jvl711:
@Narflex 

Pushed another update.  Let me know how it looks.

Josh

java/sage/VideoFrame.java
```diff
@@ -5639,20 +5640,64 @@ private boolean performLanguageMatch(String defaultLang, String[] availLangs, in
...
     targetLang = targetLang.toLowerCase();
     for (int i = 0; i < langOptions.length; i++)
     {
       if (langOptions[i].toLowerCase().indexOf(targetLang) != -1)
-        return i;
+      {
+        selectedIndex = i;
+        
+        //Prefer subtitle tracks that are not forced
+        if(langOptions[i].toLowerCase().indexOf("[" + Sage.rez("forced") + "]") != -1)
```
Narflex, line 5654: This is preferring forced subtitle tracks...the inverse of the comment.

java/sage/VideoFrame.java
Narflex: Sorry, you're right. :)

java/sage/VideoFrame.java
jvl711: Opps.  😔

jvl711:
@Narflex 

Fixed the logic mistake.

Josh

Narflex:
I cleaned up a couple import statements myself, all looks good now. :)

PR #436, opened 2020-05-02
Fix MpegDeMux DWORD alignment causing playback crash

wnjj:
This fixes an issue in MpegDeMux.ax

When storing the MPEG2 sequence header into bottom of the MPEG2VIDEOINFO data structure it is defined as an array of type DWORD.  Since the MPEG2 sequence headers are by nature 22, 86 or 150 bytes they are not naturally DWORD aligned.  Some MPEG2 files crash when the downstream filter is trying to connect with these native header sizes.  Padding the header and most critically the cbSequenceHeader member to a modulo 4 value corrects this.

The forum issue is:  https://forums.sagetv.com/forums/showthread.php?t=66336

MPEG2VIDEOINFO  definiton: https://docs.microsoft.com/en-us/previous-versions/ms787321(v%3Dvs.85)

The crash was reproduced on my system and the fix confirmed.

### DIFF
--- a/native/ax/MpegDeMux-3/CDemuxer.cpp
+++ b/native/ax/MpegDeMux-3/CDemuxer.cpp
@@ -911,15 +911,18 @@ HRESULT CDemuxer::SetMpeg2VideoMediaType(CMediaType *cmt, MPEG_VIDEO *pMpegVideo
     cmt->majortype = MEDIATYPE_Video;
     cmt->subtype = MEDIASUBTYPE_MPEG2_VIDEO;
 
+    int header_padded_length = pMpegVideo->actual_header_length + (4 - (pMpegVideo->actual_header_length % 4)) % 4; // pad length to DWORD boundary
+
     MPEG2VIDEOINFO *videoInfo = // This macro finds the pointer to the last element in the sedhdr block to determine size
         //(MPEG2VIDEOINFO*)cmt->AllocFormatBuffer(FIELD_OFFSET(MPEG2VIDEOINFO, dwSequenceHeader[pMpegVideo->lActualHeaderLen]));
 		// There's already a DWORD in the block for the sequence header, so subtract 4 bytes
-		(MPEG2VIDEOINFO*)cmt->AllocFormatBuffer(sizeof(MPEG2VIDEOINFO) + pMpegVideo->actual_header_length - 4);
+		(MPEG2VIDEOINFO*)cmt->AllocFormatBuffer(sizeof(MPEG2VIDEOINFO) + header_padded_length - 4);
     if (videoInfo == NULL) {
         return E_OUTOFMEMORY;
     }
+
 	// reset the header's memory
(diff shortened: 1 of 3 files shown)


native/ax/MpegDeMux-3/CDemuxer.cpp
```diff
@@ -911,15 +911,18 @@ HRESULT CDemuxer::SetMpeg2VideoMediaType(CMediaType *cmt, MPEG_VIDEO *pMpegVideo
     cmt->majortype = MEDIATYPE_Video;
     cmt->subtype = MEDIASUBTYPE_MPEG2_VIDEO;
 
+    int header_padded_length = pMpegVideo->actual_header_length + (4 - (pMpegVideo->actual_header_length % 4)) % 4; // pad length to DWORD boundary
```
Narflex, line 914: How about you put an ALIGN macros in the file like this and use that instead?

#define ALIGN(A, B) (((A) + (B)-1) & ~((B)-1))

Then it'd be:
int header_padded_length = ALIGN(pMpegVideo->actual_header_length, 4);

native/ax/MpegDeMux-3/CDemuxer.cpp
```diff
@@ -953,7 +956,7 @@ HRESULT CDemuxer::SetMpeg2VideoMediaType(CMediaType *cmt, MPEG_VIDEO *pMpegVideo
 	
 	videoInfo->hdr.dwInterlaceFlags = pMpegVideo->progressive ? 0 : (AMINTERLACE_IsInterlaced);
 	videoInfo->hdr.AvgTimePerFrame = pMpegVideo->picture_time;
-    videoInfo->cbSequenceHeader = pMpegVideo->actual_header_length;
+    videoInfo->cbSequenceHeader = header_padded_length;
```
Narflex, line 959: I agree with allocating more...but should you be setting the size to the aligned size and not the actual size?  Did you test if that works OK?

wnjj: I thought you would notice that.  ;)  I suppose I should have included that in my first comment.

All of my testing showed that the cbSequenceHeader was really what needed to be DWORD aligned.  It still crashed even when I allocated and zeroed lots of extra buffer size but left cbSequenceHeader alone.  There must be something about accessing or freeing the memory that makes use of cbSequenceHeader.  Since the header information is formatted and parsed I imagine it wouldn't matter if there are pad bytes after it.

The commit I submitted was the one I tested.

Narflex: OK, that's good enough for me. :)  If you wouldn't mind making those statements cleaner with an ALIGN macro, that'd be great and then I'll approve it for merge.

wnjj: Done.  I named it ALIGN_DWORD since it's specific to a size of 4.

PR #442, opened 2020-09-02
Fix conversion to MKV

jvl711:
There was an inadvertent consequence to adding a format substitution from "MATROSKA,WEBM" -> "MATROSKA".  This caused conversions to MKV container to break.

For more info see https://forums.sagetv.com/forums/showthread.php?t=66492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## Next
+* Fix for conversion to MKV. Removed Format substitution of "MATROSKA" -> "MATROSKA,WEBM"
+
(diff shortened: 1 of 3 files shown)


Narflex:
Thanks for the fix. This will break this for the plugin format detector, but it can get fixed within that plugin itself since this breaks things in Sage TV otherwise.

jvl711:
@Narflex 

Thank you for handling this quickly.  I will have an update for the format parser in the next week or so.  FYI, I have another pull request coming soon for some enhancements to fixed transcoding.  Giving a little more time for some beta testing feedback.

Thanks,
Josh

PR #445, opened 2020-09-15
Transcoding enhancements no debug

jvl711:
I made some minor changes to the behavior of how SageTV determines when to use fixed transcoding.  I added a number of properties to the fixed transcode section to allow for HD transcoding.  I tried my best to make the changes backwards compatible.  I do not believe they will interfere with any of the other clients.  There are a few users using this, and it appears to be working well.

Josh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 
 ##Next
-* Added ability to used fixed push format when transcoding is required, but not low bandwidth
+* Added ability to use fixed push format when transcoding is required, but not low bandwidth
 * Added 720, 1080 and SOURCE (use video source resolution) options to FFMPEGTranscoder for fixed transcoding
 * Added SOURCE option to FPS that calculates GOP automatically and uses FPS of source videos
 * Added an option for Audio Channels.  Does not allow a value greater than source audio
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+##Next
+* Added ability to used fixed push format when transcoding is required, but not low bandwidth
+* Added 720, 1080 and SOURCE (use video source resolution) options to FFMPEGTranscoder for fixed transcoding
+* Added SOURCE option to FPS that calculates GOP automatically and uses FPS of source videos
+* Added an option for Audio Channels.  Does not allow a value greater than source audio
+
 ## Version 9.2.2 (2020-05-16)
 * Fixed MpegDeMux that crashed some MPEG2 playback (Windows)
 * Change service launcher (Windows) to support local JRE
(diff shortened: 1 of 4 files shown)


Narflex:
You'll need to resolve the conflict like it says before I can merge changes.

java/sage/MiniPlayer.java
```diff
@@ -1097,6 +1097,12 @@ else if (!containerOK)
             sage.media.format.ContainerFormat currFileFormat = currMF.getFileFormat();
             if (currFileFormat != null && "true".equals(currFileFormat.getMetadataProperty("VARIED_FORMAT")))
               currFileFormat = sage.media.format.FormatParser.getFileFormat(file);
+            //Check to see if there was a fixed format defined for transcoding
```
Narflex, line 1100: Why did you make this change?

Narflex (changes requested):

java/sage/MiniPlayer.java
jvl711: There were cases in my testing where it would fall through the checks and not get set to the fixed format.  It would use the default defined in some cases.  I wanted to use fixed format when the codec/container combination is not supported.  I added an option to the Android client where it is fixed format, but passes the supported codecs.  I think the Placeshifter sets all the Pull formats to null when it is in Fixed mode.  Let me know if you want specific scenarios.  I can go thru and retest without it and see where it fails.  You may have a better suggestion to fix it than what I did there.  I figured it was a low risk change because we have already decided we are going to use the transcoder, and if we have a fixed format set, I would think we should be using it.

Thanks,
Josh

Narflex: Yeah, this code is really tricky and there's plenty of times I thought I needed changes like that but they broke other scenarios.  Definitely retest it again w/out this change and then figure out what the actual problem is so we can fix it properly w/out breaking other things.

jvl711: Here is a link to a heavily debug logged output of a reproduction of the issue.
[https://drive.google.com/file/d/15qDOVIBmNMVC8wNF2lOQNEDw9BoU-Nhv/view?usp=sharing](url)

Here are the conditions.

1. Android Client
2. Local Playback
3. Playing LiveTV of MPEG2
4. Android Client sent a list of supported pull codecs/formats.  MPEG2 is not supported
5. Android Client sent a fixed format to SageTV Server
6. SageTV server decided to use DVD6CH instead of the fixed format

I am not sure which condition is the wrong one.  Maybe my assumption is wrong on the outcome.  Let me know what you think.  Let me know if you have an issues accessing the log, or need additional output.

Josh

Narflex: OK, the reason it wasn't working is because the AndroidClient acts like a media extender...but it also has the ability to set a fixed push format (which our extenders never did).  So overall this change does seem reasonable...with one exception...I think this'll break if you then try to use the AndroidClient as a placeshifter and try to play back audio and it transcodes it. So can you put another condition in there so this only gets triggered if the file has video content?

jvl711: I made a change to make sure there is more than one video format in the file.  I think this should work.  Let me know if you see anything else.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+##Next
+* Added ability to used fixed push format when transcoding is required, but not low bandwidth
```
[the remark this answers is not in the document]
jvl711: I merged the files.  I did not know that I could do that :)

jvl711:
I think I resolved all of the issues.  Let me know if I missed anything.

Josh

Narflex:
Looks good now, thanks for the changes!

PR #454, opened 2021-07-28
Added option to FFMpegTranscoder to be able to copy audio or video without conversion (remux)

jvl711:
This is the first of a few changes to allow for remuxing when the container format is not supported using the miniclient.  I am going to add a property for a fixed remux format.

Let me know if you have any questions or concerns.  High level testing has looked good.  Submitting in smaller pieces.

### DIFF
--- a/java/sage/FFMPEGTranscoder.java
+++ b/java/sage/FFMPEGTranscoder.java
@@ -644,7 +644,7 @@ public class FFMPEGTranscoder implements TranscodeEngine
         }
         else
         {
-          xcodeParams += " -vcodec " + vcodec + " -s " + s + " -ac " + ac + " -g " + g + " -bf " + bf + (deinterlace ? " -deinterlace " : "");
+          xcodeParams += " -vcodec " + vcodec  + " -b " + b + " -r " + r + " -s " + s  + " -g " + g + " -bf " + bf + (deinterlace ? " -deinterlace " : "");
         }
         
         if(acodec.equals("COPY"))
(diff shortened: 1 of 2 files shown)


Narflex:
I'm curious why you did this rather than just leaving it the way it was and using "copy" for the codec name. This is really just adding the ability to use "COPY" as well as "copy" to achieve the result.

jvl711:
I did not want any of the other command line options being sent when copying <shrug>.  I assume it might work, but seems kinda ugly if you passed unused/unneeded .  I have not tested it, FFMpeg may complain.

I am fine changing it if you prefer it designed different!

java/sage/FFMPEGTranscoder.java
```diff
@@ -635,9 +635,29 @@ else if ("container".equals(propName))
...
+        
+        xcodeParams = "-f " + f;
+        
+        if(vcodec.equals("COPY"))
+        {
+          xcodeParams += " -vcodec copy";
+        }
+        else
+        {
+          xcodeParams += " -vcodec " + vcodec + " -s " + s + " -ac " + ac + " -g " + g + " -bf " + bf + (deinterlace ? " -deinterlace " : "");
```
Narflex, line 647: "ac" is audio channels, so that should go with the audio codec. "b" is video bitrate, so that should go with video. "r" is the frame rate, so that should go with video.

Narflex:
I looked at this too soon after I woke up and didn't realize it was doing more than what I said. Ignore my initial comment. :) But do pay attention to the review comment, there's a couple args you need to move around.  And the overall order can be changed, just be sure you keep the acodec/vcodec args before their related audio/video parameters.

java/sage/FFMPEGTranscoder.java
jvl711: Sorry about that.   I am glad that you caught that.  I only tested with audio/video copy together.  I did not test with a copy on  only audio or video.  I think I got it correct with the latest commit.

Narflex:
For some reason this didn't kick off a travis CI run...but it's good to merge.

PR #455, opened 2021-08-02
Modified the Minicliet to have a setting for fixed remuxing

jvl711:
Modified the Minicliet to have a setting for fixed remuxing when the video codec and audio codec are both supported, but the container is not.  This allows for lower cost transcoding in certain situations, and can also be used to fix files which may have issues with their containers.

### DIFF
--- a/java/sage/MiniClientSageRenderer.java
+++ b/java/sage/MiniClientSageRenderer.java
@@ -3849,6 +3849,7 @@ public class MiniClientSageRenderer extends SageRenderer
         sendGetPropertyAsync("PUSH_AV_CONTAINERS");
         sendGetPropertyAsync("STREAMING_PROTOCOLS");
         sendGetPropertyAsync("FIXED_PUSH_MEDIA_FORMAT");
+        sendGetPropertyAsync("FIXED_PUSH_REMUX_FORMAT");
         sendGetPropertyAsync("OPENURL_INIT");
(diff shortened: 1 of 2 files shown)


Narflex:
I'm wondering how useful this is. Given that this is only for push mode remuxing...push mode only operates with MPEG2-TS/PS container formats.  Are you seeing cases where a client only supports PS and not TS (or vice versa) which would then provide a benefit to remuxing like this?

What did you use as a test for this?

jvl711:
There are a few containers that are not supported on Andriod.  Specifically AVI.  In that case this is a low cost way to support those files without having to transcode.  Plus I want to be able to make a setting where all content is remuxed.  By passing no supported pull containers I can force Sage to remux all files.  This can fix some potential compatibility issues.  Remuxing is way less expensive than transcoding.

I agree that it's use will be edge cases, but I think it is a nice addition to the capability set, and can allow for an low cost workaround to certain unsupported content.  I have a proof of concept work with the Android Client where it is remuxing AVIs with very low CPU usage.  It seems to be working well.

Narflex:
Since this won't be on forcibly unless a client requests it, it should be OK. I know I didn't do this before because using codec copy in remuxing quite often can cause A/V sync issues...and not all codecs in one container can be put it another container (but that one you likely won't run into much).

PR #457, opened 2021-08-16
Added some additional constraints on -aspect swith in FFMPEGTranscode…

jvl711:
Added some additional constraints on -aspect swith in FFMPEGTranscoder to make sure an invalid aspect ratio is not passed to the transcoder.  In my testing, I saw a few occurrences when watching live TV where the content did not seem to be properly identified, and the container format had no aspect or dimensions to the video.  This caused FFMPEGTranscoder to pass -aspect 0:0.  This caused the transcoder to fail and Sage client to display a blank screen.

I realize this is a code path that should probably not occur, but I also think it is a reasonable check and restriction.  I would rather FFmpeg try again to determine the aspect when it has more data.

Josh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Change Log
 
 ## Next ##
-* Added option to FFMPEGTranscoder to allow for a setting to copy video or audio 
+* Added option to FFMPEGTranscoder to allow for a setting to copy video or audio
+* Added a new option to the Miniclient for fixed remux profile.  This is used when the audio/video codec are supported, but the container is not
+* Added some additional constraints on -aspect swith in FFMPEGTranscoder to make sure an invalid aspect ratio is not passed to the transcoder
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Change Log
 
 ## Next ##
-* Added option to FFMPEGTranscoder to allow for a setting to copy video or audio 
+* Added option to FFMPEGTranscoder to allow for a setting to copy video or audio
+* Added a new option to the Miniclient for fixed remux profile.  This is used when the audio/video codec are supported, but the container is not
+* Added some additional constraints on -aspect swith in FFMPEGTranscoder to make sure an invalid aspect ratio is not passed to the transcoder
```
Narflex, line 6: Misspelled 'switch'

jvl711:
Thanks Jeff!

PR #459, opened 2021-09-02
Create GitHub Action for CI

Narflex:


Narflex:
Yeah! We have CI again. :)  Still needs to be updated to do deploy...and also increase the Ubuntu version since 16.04 is going to be deprecated this month by GitHub Actions.

PR #460, opened 2021-09-02
Rename file and update Ubuntu version for CI

Narflex:
I named this file wrong, fix that. Also I need to see if it works with a newer version of Ubuntu. GitHub Actions is deprecating 16.04 this month.


Narflex:
Awesome, CI works again. :)

PR #461, opened 2021-09-02
Fix bool defs in Channel.h

Narflex:
These just look wrong and have caused various problems, if they break anything they can be put back.

### DIFF
--- a/native/ax/Channel-2/Channel.h
+++ b/native/ax/Channel-2/Channel.h
@@ -54,16 +54,6 @@
 
 #define MAX_SATELLATE_NUM  6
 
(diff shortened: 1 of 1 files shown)


Narflex:
I should mention this was causing the local build on my workstation to fail as well as the 18.04 CI build.

wnjj:
This did break native compiling on Windows.  I'm trying to see what it takes for bool to be natively recognized but it's possible other parts of the code are defining it to nothing.  Can I put it back but under the #ifdef WIN32?  That should only affect Windows compiling.

Narflex:
Merge accepted.

PR #462, opened 2021-09-13
Add GitHub Action for deploy and increment version

Narflex:


Narflex:
I'm not entirely sure this will cause the deploy action to work. My concern is that the build output artifacts from the 'build' step won't be usable by the 'deploy' step...but we will find out when I push it.

PR #465, opened 2021-09-29
Revert PR#461 for Windows build

wnjj:
bool, true and false are not part of WIN32 natively.  This reverts https://github.com/google/sagetv/pull/461 but only for Windows compiling.


Narflex:
OK, not surprising...I figured we left it there for some reason. Thanks for fixing!

Issue #468, opened 2021-12-30
Difference in Sage.jar between Windows and Linux builds


jusjoken:
I have been working on rebuilding the Unraid Docker solution and was running into an issue with the Java8 docker as SageTV would run but a placeshifter would not connect.  The issue was related to a change in Java between 8 and 9 and Byte(ByteBuffer).

The odd thing is I often test using Java 8 on windows and had no issues connection a placeshifter...both builds were the latest 9.2.6 releases.

So I copied the Sage.jar from the windows release into the linux docker and restarted SageTV and now I was able to connect the placeshifter to this Java 8 docker.

In my research I find that you can build a jar using JDK9 with the --release=8 option and it will resolve the Byte/ByteBuffer compatibility issue.

So are we using this --release option for the Windows jar build and not the Linux build or what is different in the 2 build processes that makes the Linux build fail on Java 8 but the Windows build works?

Thoughts?

enternoescape:
I obviously cannot speak for everyone, but this would allow the older docker container to work with Java 8 again which would address some compatibility issues with older plugins. On a side note, when I was building to fix an issue on Windows, I had to upgrade to Java 9 to be able to successfully compile the current code as it is on Github. I wasn't expecting that and spent a lot of time troubleshooting assuming I created the issue.

wnjj:
I’m pretty sure I built the last windows version with JDK 8 which would explain the difference. JDK9 was a result of needing to update the distribution for the automated build on GitHub.

jusjoken:
I currently do not have a build system for windows nor Linux but as time permits I will set one up again.  If someone doesn't beat me to it I will test out the release=8 option under jdk9 and see what happens.

I think if we can support Java 8 as long as possible the better as we have many older in use plugins that could be affected.   The Jetty issue may be the biggest but I am close to a jetty 9 release that works well on Java 8+

jusjoken:
I found a resolution to this.  I could not make it work with a build system that uses JDK9 as the --release option under Gradle is broken and Gradle does not plan to resolve it.  So I built a build system as follows...

- JDK 11 and ubuntu 20 (I actually used "adoptopenjdk/openjdk11:jdk-11.0.11_9" docker as my base image).  
- Gradle 7.3.3 (needed as it supports --release)
- Had to change the gradle.build a bit to support Gradle 7+ as compile has been removed and now needed to use implementation

[...]

PR #469, opened 2022-01-02
Changes to support a linux build to run on Java 8

jusjoken:
Resolves issue https://github.com/google/sagetv/issues/468

Still requires changes to the github workflow for automatically building linux bilds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,14 +10,14 @@ jobs:
   build:
 
 #    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


jusjoken:
As info - the automated build process fails as it uses JDK 9 which under Gradle will fail (known bug they will not fix) .  I had to use JDK 11

jusjoken:
I did some research and made the appropriate changes to the workflow to now run under ubuntu 20.04 and use JDK 11 and the build passes

Narflex:
Thanks for fixing this!

PR #470, opened 2022-01-04
Update changelog [ci release]

jusjoken:


jusjoken:
The deploy was skipped on the Google workflow but the deploy ran and failed on my jusjoken github account.  It failed stating opensagetv token not set.  Is this how it is supposed to deploy...from my account and if so where would I set the token?

Narflex:
Yeah, it won't run on your Google workflow. It'll run after I do the merge commit. It indicated it succeeded here: https://github.com/google/sagetv/runs/4706333747?check_suite_focus=true

jusjoken:
I see that.  I guess I didn't wait long enough. Interesting that it does run on my GitHub but fails.  Wonder if I need to change a setting so it doesn't run on the fork or just ignore it as it fails anyway

PR #471, opened 2022-01-05
Update build process to support ubuntu 18.04 as it is still under support [ci release]

jusjoken:
As the ubuntu 20.04 build forces glibc to be 2.28 the linux SageTV build would not run on Ubuntu 18.04 which is still under support until mid 2023.  Given that 18.04 may still be used by users for awhile I updated the build process to use Ubuntu 18.04 which allows SageTV build to run on either 18.04 and 20.04 (and their equivalents) .
I also renamed the travis.yml so it would quit running and failing on forked repos

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
   build:
 
 #    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


Narflex:
Thanks again!

PR #480, opened 2022-08-13
Update transcoder progress to fallback to framecount

jvl711:
In certain situations ffmpeg does not output a correct time in the status output.  It seems like this happens most often when doing a remux of the file instead of a transcode.  When the time is calculated as < 0 I am falling back to calculating the progress based on framecount.

I also made some minor changes in the gradle file to allow this project to build under netbeans.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## Next
+* Updated grable script so that project could build in Netbeans
```
Narflex, line 4: s/grable/gradle

java/sage/FFMPEGTranscodeJob.java
```diff
@@ -120,6 +120,7 @@ public float getPercentComplete()
     if (tempy != null)
     {
       long fullLength = clipDuration == 0 ? (mf.getDuration(transcodeSegment) - clipStartTime) : clipDuration;
+
```
Narflex, line 123: Can you undo this line removal?

build.gradle
```diff
@@ -69,7 +69,7 @@ sourceCompatibility = 1.7
 
 compileJava {
   options.warnings = false
-  options.compilerArgs.addAll(['--release', '8'])
+  //options.compilerArgs.addAll(['--release', '8'])
```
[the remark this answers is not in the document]
jvl711: I am going to put this back.  I took it out because I was not able to build at all with it in.  I am getting this error with it in "invalid flag: --release".  I am wondering if adding this is necessary if we are doing targetCompatibility = 1.7 and sourceCompatibility = 1.7.

build.gradle
```diff
@@ -315,10 +315,13 @@ task(packageSageJar)  {
  *
  * @return build number
  */
-def getBuildNumber() {
-  def buildVerText = new File('java/sage/SageConstants.java').text
+def getBuildNumber()
```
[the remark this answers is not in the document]
jvl711: This was intentional.  I think it is actually incorrect today.  When you use "new File()" it's relative path is not always the project root.  "file()" on the other hand constructs a groovy file object, with the relative path of the file being the project root.  When I tried to build the project from Netbeans it failed without this change.  It is not able to find the file.  I believe the change I made is a bug fix in the script.

java/sage/FFMPEGTranscoder.java
```diff
@@ -1363,24 +1363,52 @@ else if (c == '\r')
...
+                    {
+                      if (XCODE_DEBUG) System.out.println("Using framecount to calculate transcoder progress");
+                      
+                      if(frameIdx != -1)
+                      {
+                        frameStr = sb.substring(frameIdx + 6, fpsIdx).trim();
+                      }
+                      
+                      int frame = Integer.parseInt(frameStr);
+                      float fps = FFMPEGTranscoder.this.sourceFormat.getVideoFormat().getFps();
```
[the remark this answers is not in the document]
jvl711: I did some testing since the FFmpeg documentation was not specific.  I am fairly certain it is the output format that dictates the frame count.  It does not look like we have a container format for what format we are going to.  Do you have a recommendation for getting the target fps?

build.gradle
```diff
@@ -315,10 +315,13 @@ task(packageSageJar)  {
  *
  * @return build number
  */
-def getBuildNumber() {
-  def buildVerText = new File('java/sage/SageConstants.java').text
+def getBuildNumber() 
+{
```
Narflex, line 319: Put this brace back on the prior line (it's an unnecessary change).

build.gradle
```diff
@@ -315,10 +315,13 @@ task(packageSageJar)  {
...
+  def buildVer
```
Narflex, line 320: Erase this line and the next line and put the 'def' right before buildVer a coouple lines down.

build.gradle
```diff
@@ -355,10 +358,14 @@ def getBuildNumber() {
  */
 task(updateBuildNumber)  {
   doLast {
+
```
Narflex, line 361: Remove these extraneous blank lines that were added.

java/sage/SageConstants.java
```diff
@@ -22,5 +22,5 @@ private SageConstants()
     // Non-instantiable
   }
 
-  public static final int BUILD_VERSION = 441;
+  public static final int BUILD_VERSION = 999;
```
Narflex, line 25: Don't modify this file.

java/sage/FFMPEGTranscoder.java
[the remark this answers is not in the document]
Narflex: Aside from tracking what we launch...not really. From what I can tell...the only impact this change will have is on the progress duration showed in the UI, so probably not worth more effort. Is there anything else you noticed it would actually fix?

Narflex:
Thanks for the update!

PR #481, opened 2022-09-17
* Fix Subtitles on Extender only using 1st subtitle

JREkiwi:
On extenders irrespective of the subtitle selected, the extender will only ever play the 1st subtitle.

Logs show
Subtitle change is done; subtitle is: # 3 eng
Enabling subpicture stream 0

Added some debug to identify where the disconnect was and the issue resolved. Removing the debug and the issue returns

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ## Next
+* Fix Subtitles on Extender only using 1st subtitle irrespective of which subtitle is selected
 * Updated gradle script so that project could build in Netbeans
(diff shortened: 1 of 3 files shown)


JREkiwi:
I don't understand why this fixes it.

Narflex:
I don't think this necessarily is fixing anything...it's just debug lines. When that *does* fix something, it's usually because of some other timing issue that's involved that the debug logging slows down (or there's something else wrong with the testing being done).

JREkiwi:
You are right.  It's an issue with the MediaFormatParserPlugin setting an id=0 for all subtitle streams.

PR #493, opened 2025-03-08
SD EPG fixes [ci release]

jusjoken:
Various commits to fix the SD EPG issues


java/sage/SageConstants.java
Narflex, line 1: Why are you deleting this file?

java/sage/EPG.java
```diff
@@ -597,6 +598,12 @@ public java.io.File getLogoDir()
 
   public void run()
   {
+
+    //03-02-2025 jusjoken: added a version check here.  As this EPG process runs basically daily I added the
+    //  version check here.  It may fit best elsewhere but I will leave it here for now to ensure future updates are noticed
+    //check if an updated version is available on github
+    wiz.checkForUpdate();
```
Narflex, line 605: You should probably remove this, and just leave the one below that happens in the loop. It's generally not a great idea to do some kind of network check during startup unless it's critical (and this is just informative where it creates a system message).

java/sage/EPG.java
```diff
@@ -624,22 +631,47 @@ public void run()
...
     boolean updateFinished = true;
+    //JUSJOKEN: 2025-02-19 add ability to run FULL Maintenance at a specific time each day
     while (alive)
     {
+
       try{
 
-        if (Sage.time() - wiz.getLastMaintenance() > MAINTENANCE_FREQ)
-          reqMaintenanceType = MaintenanceType.FULL;
+        if(Sage.getBoolean("wizard/scheduled_maintenance", false)){
```
Narflex, line 640: It looks like you're disabling Wizard maintenance here unless this new property is set (which it will not be by default of course). That's not good, as that will lead to memory bloat if the DB doesn't run it's regular maintenance routine (and users will run out of memory sooner than later without DB maintenance). You may be misunderstanding what 'maintenance' is here...and it's when the Wizard runs through the database and clears out the old unneeded data...it's orthogonal to EPG updates.

java/sage/EPG.java
```diff
@@ -624,22 +631,47 @@ public void run()
...
+              reqMaintenanceType = MaintenanceType.FULL;
+            }
+        }
 
         if (reqMaintenanceType != MaintenanceType.NONE)
         {
           Carny.getInstance().kickHard();
           SchedulerSelector.getInstance().kick(false);
         }
+
```
Narflex, line 665: All these code additions which are just commented out debug lines should be removed.

Narflex (commented):
Overall good, please address the comments and explain the maintenance issue in the EPG class. (note to self, I reviewed everything completely except EPG.java)

Narflex (approved):

Narflex:
That "approve changes" you see was just for a single commit (and maybe more after that)...I forgot the review applies to the whole PR. :)

Narflex (commented):
Fully approve "SD EPG Updates - added seeker/duration_for_watchdog to allow users to…"

Narflex (commented):
Fully approve: "Github actions - changed to 20.04 as 18.04 is no longer supported"

Narflex (commented):
The commit for "Revert "Add properties to allow changing locator server"" shouldn't be in here...you likely need to do a local 'git pull' from the sagetv repo and then rebase your local branch on top of that...then it should remove that commit from your history.

build.gradle
Narflex, line 1: Why did you make these changes to build.gradle?

Narflex (commented):
You should squash this commit w/ the other one for the changes to build.gradle, then at least the Linux versioning portion will be more clear.

Narflex (approved):
"Release for version 9.2.9 [ci release]" is approved

java/sage/epg/sd/SDUtils.java
```diff
@@ -409,6 +434,15 @@ public static String fromSageTVtoProgram(String program)
       program.getChars(2, 12, returnValue, 4);
       return new String(returnValue);
     }
+    if (program.length() == 10 && program.startsWith("EP"))
```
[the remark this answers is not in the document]
jusjoken: Schedules Direct (SD) updated their api to require all program IDs to be 14 characters.  They have temporary processes in place to fix any SageTV programID that is less than 14 so they asked me to fix it on our side.

java/sage/SageConstants.java
jusjoken: Sorry.  This must have occurred by mistake during my squashing as their was a conflict.  I will add it back in in another commit.

java/sage/EPG.java
jusjoken: Ok.  Done and will be in a new commit

java/sage/EPG.java
jusjoken: Not my intention here. This is still running maintenance but at a specific hour every day.  If this property is set to true it still sets maintenance to FULL but the timing is based on the scheduled_maintenance_offset (hour of day) having passed.  Hope that makes sense.

java/sage/EPG.java
jusjoken: Done in next commit

build.gradle
jusjoken: The ant.exec command function no longer works in ubuntu 22.04 so I had to change it to the executable function that replaced it.

java/sage/EPG.java
Narflex: OK, now I see...I missed the 'else' on this conditional below before. :) However, to avoid confusion...can you rename the 'EPG update offset' to be something like that...such as wizard/scheduled_epg_update?  Maintenance in SageTV does not refer to EPG updates...it refers to when the DB is analyzed and all expired records are removed to reduce memory usage.

java/sage/EPG.java
```diff
@@ -624,19 +626,37 @@ public void run()
...
-          reqMaintenanceType = MaintenanceType.FULL;
+        if(Sage.getBoolean("wizard/scheduled_maintenance", false)){
+            if (Sage.DBG) System.out.println("EPG next scheduled maintenance offset is " + Sage.getInt("wizard/" + SCHEDULED_MAINTENANCE_OFFSET, 0));
+
+            nextScheuldedMaintenanceTime = getNextScheduledMaintenanceTime();
+
+            if ((Sage.time() - wiz.getLastMaintenance() > MAINTENANCE_FREQ) || ((nextScheuldedMaintenanceTime - MAINTENANCE_FREQ) < Sage.time())){
+                if (Sage.DBG) System.out.println("EPG next scheduled update is ready to run as we are past the maintenance window and/or scheduled time");
+                reqMaintenanceType = MaintenanceType.FULL;
+            }else if(wiz.getLastMaintenance()==0){
```
Narflex, line 643: I don't think you need this 'else if' block here, because of wiz.getLastMaintenance() is zero...then the first part of the 'if' would be triggered...so this block will never be executed.

java/sage/EPG.java
```diff
@@ -624,19 +626,37 @@ public void run()
...
     {
+
       try{
 
-        if (Sage.time() - wiz.getLastMaintenance() > MAINTENANCE_FREQ)
-          reqMaintenanceType = MaintenanceType.FULL;
+        if(Sage.getBoolean("wizard/scheduled_maintenance", false)){
+            if (Sage.DBG) System.out.println("EPG next scheduled maintenance offset is " + Sage.getInt("wizard/" + SCHEDULED_MAINTENANCE_OFFSET, 0));
+
+            nextScheuldedMaintenanceTime = getNextScheduledMaintenanceTime();
```
Narflex, line 638: nextScheuldedMaintenanceTime is misspelled

java/sage/EPG.java
```diff
@@ -651,161 +671,189 @@ public void run()
...
+
         synchronized (sources)
         {
           for (int i = 0; (i < sources.size()) && alive; i++)
           {
+            if(Sage.getBoolean("wizard/scheduled_maintenance", false)){
+                //determine the wait until the next scheduled update time
+                //calc the next time as the user may have changed the schedule settings
+                nextScheuldedMaintenanceTime = getNextScheduledMaintenanceTime();
+                minWait = nextScheuldedMaintenanceTime - Sage.time();
```
Narflex, line 683: Isn't there a problem where if you set minWait in this block...then it can get overridden on line 687 with the one from the EPGDataSource?  So the scheduled maintenance only takes effect if the last EPGDataSource has an update time further out then the scheduled time.  It seems like this check should be outside of this loop to override the other update timing logic.

build.gradle
```diff
@@ -406,7 +409,7 @@ compileJava.dependsOn updateBuildNumber
 sageJar.dependsOn cleanSageJar, classes, test
 miniclientJar.dependsOn clean, miniclientClasses, cleanMiniJar
 linuxMiniClientRelease.dependsOn miniclientJar
-sageJar.finalizedBy restoreSageConstants
```
Narflex, line 409: Why was this line removed?

java/sage/EPG.java
jusjoken: Agreed.  Removed (I believe I added while testing so I could force an update on restart by setting it to 0 but should have removed this).

java/sage/EPG.java
jusjoken: fixed

Narflex:
All the latest changes look good. I just want to build it locally and install it on my server now for a couple days to test it. :) I'll also trigger the CI runs as well. If all goes well, then I'll merge it in a few days.

java/sage/EPG.java
andyvshr: BTW: Will this new epg scheduled update control affect all types of epg updates, or only Schedules Direct?

Narflex:
Just figured out my local build issue (wrong version of Java) so now I have it built and will also test it...now that it's all been merged. :)

java/sage/EPG.java
Narflex: It will affect all types of EPG updates...but only if the user has enabled those options.

PR #494, opened 2025-03-08
Add some EPG options to the UI

andyvshr:
STV v9.2.9.2 2025030701
Added these options:
Detailed Setup -> General
a) Enable/Disable - Automatically check GitHub for SageTV Update Releases 
Detailed Setup -> Customize
b) Enable use of Program images from Schedules Direct: Yes/No 
c) Enable use of Celebrity images from Schedules Direct: Yes/No
d) Enable Schedules Direct EPG Updates: Yes/No - Disabled
e) Schedules Direct EPG Update Hour: set an hour of the day to update or let it be automatic


Narflex:
Andy, I shouldn't have merged this yet because I had him change the property names...so they no longer match what is in the STV. Can you fix it so they match now and then I'll merge the new one?

andyvshr:
I saw - was planning to change the STV either today or tomorrow. Might be too late to start some caulking today, so maybe I'll do this now.

PR #499, opened 2025-05-16
Fix for return of old token

jusjoken:


Narflex:
Can you rebase this back on the master branch? I fixed the problem with the failing GitHub actions and want to re-run this against it before merging.

jusjoken:
I believe I have updated as you requested

PR #505, opened 2025-06-17
Fix daily notification of new version.  Should only notify once when …

jusjoken:
…new version is available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+## Version 9.2.14 (2025-06-16)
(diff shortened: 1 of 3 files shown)


Narflex:
Remember...I retired from Google...so I can't approve pull requests in this repo anymore. You should sync the OpenSageTV one to this one and then do the pull request there.  I could get us going again on this one (some SageTV employees are at Google still) if we just wanted to get this through since there likely won't be any further changes for awhile after this. Let me know which route you want to go.

Narflex (approved):

PR #507, opened 2025-09-16
Fixes for SD

jusjoken:
Fix for lockouts with SD EPG including better handling when run as service and using clients. 
Also ensure SD is sent 14 char program ids

### DIFF
--- a/java/sage/MetaImage.java
+++ b/java/sage/MetaImage.java
@@ -3231,7 +3231,9 @@ public class MetaImage
                 if (Sage.DBG) System.out.println("MetaImage.loadCacheFile: Found SD image url. src = '" + src + "'");
 
                 //skip the image loading if the bypass properties are set
-                if(Sage.getBoolean("sdepg_core/bypassCelebrityImages", false) && Sage.getBoolean("sdepg_core/bypassProgramImages", false)){
+                //2025-09-18 to ensure the server property is used even on a client SageProperties.java was updated to transfer these properties from the server
+                //  - as we don't know the image is Celebrity vs Program then skip loading for either setting to be safe
+                if(Sage.getBoolean("sdepg_core/bypassCelebrityImages", false) || Sage.getBoolean("sdepg_core/bypassProgramImages", false)){
                     if (Sage.DBG) System.out.println("MetaImage.loadCacheFile: skipping image load as both bypass properties are set. src = '" + src + "'");
                     break;
                 }
--- a/java/sage/epg/sd/SDSession.java
+++ b/java/sage/epg/sd/SDSession.java
@@ -58,6 +58,8 @@ import java.security.InvalidParameterException;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Collection;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public abstract class SDSession
 {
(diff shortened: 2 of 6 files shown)


java/sage/MetaImage.java
```diff
@@ -3230,9 +3230,36 @@ private boolean loadCacheFile()
                 //this is an SD supplied image so a token is required to retreive it
                 if (Sage.DBG) System.out.println("MetaImage.loadCacheFile: Found SD image url. src = '" + src + "'");
 
-                //skip the image loading if the bypass properties are set
-                if(Sage.getBoolean("sdepg_core/bypassCelebrityImages", false) && Sage.getBoolean("sdepg_core/bypassProgramImages", false)){
-                    if (Sage.DBG) System.out.println("MetaImage.loadCacheFile: skipping image load as both bypass properties are set. src = '" + src + "'");
+                //get the bypass properties
```
Narflex: In SageProperties.java, there's a list there of properties that get sent from the server down to the client, that's the better way to do this rather than invoking server API calls every time the client wants to load an SD image.  Then you can just retrieve the property locally w/ no performance hit at all.  I know this same GetServerProperty logic appears elsewhere in SD code, but you can leave those alone.

Narflex (changes requested):

java/sage/epg/sd/SDSession.java
```diff
@@ -213,12 +215,21 @@ public String getUsername()
...
-   * Returns the provided token.
+   * Returns the current valid token.
    *
-   * @return The current token.
+   * @return The current valid token.
    */
-  public synchronized String getToken()
+  public synchronized String getToken() throws IOException, SDException
   {
+    if (System.currentTimeMillis()/1000 < tokenExpiration && token != null)
```
Narflex: It looks like you can just remove this conditional block and always call authenticate() because the first thing that method does is this same check and immediately return if it's true....so having it here as well is just duplicating code.

java/sage/MetaImage.java
jusjoken: I may need some guidance here.  I see 2 lists in this source and the 2 are nearly identical in content other than 2 entries...

  public static final String[] CLIENT_XFER_PROPERTY_PREFIXES = { 
    "mmc/video_format_code",
  
  public static final String[] CLIENT_DONT_SAVE_PROPERTY_PREFIXES = { 
    "mmc/encoders", 

All the other entries in these list exist in both lists. So how are these 2 lists used and do I add...

sdepg_core/bypassCelebrityImages
and
sdepg_core/bypassProgramImages

to 1 or both lists.

Narflex: Yeah, I'm curious what I was thinking 20 years ago with some of this...I could have at least commented my own code better...guess I'lll have to dig around. :)  So the purpose of the 2 lists at a high level should be somewhat clear. One is for properties that get transferred from the server to the client upon connection. The other is for properties the client shouldn't write out to disk.

The reason for the 2 outliers you found are because 'mmc/encoders' should never actually appear on the client...and there was probably some legacy bug at some point where it did and it caused a problem, so this was an easy way to prevent that from happening again.  The 'mmc/video_format_code' is saved on the client because there's a couple UI paths which do different window sizes based on NTSC vs. PAL video formats, and these windows are created before a server connection is established...I'm not sure that affects current code anymore, but it was used to at some point, so it's retained on the client for that reason.

As for the current change, put them in both lists since there's no reason for them to have special treatment. Thanks!

jusjoken: updated

java/sage/epg/sd/SDSession.java
jusjoken: done

PR #514, opened 2026-04-06
Security: Fix command injection in NFS mount check and CD burn status

r3352:
## Summary

- Two code paths pass configuration property values into `sh -c` shell commands, enabling command injection via `SetServerProperty` API
- Replaces shell-based execution with safe alternatives that don't interpret shell metacharacters

## Vulnerability Details

**CWE-78** (OS Command Injection)

Two code paths use `exec2(new String[]{"sh", "-c", ...})` with unsanitized property values concatenated into the command string:

1. **`IOUtils.java:1147`** — NFS mount check: `"mount -t nfs | grep -i \"" + localPath + "\""` — the `localPath` from `linux/nfs_mounts` property can escape the double quotes
2. **`MediaFile.java:3688`** — CD burn status: `burnPath + " -toc dev=" + devPath` — both `linux/cd_burn` and `default_burner_device` properties are concatenated directly

Since `SetServerProperty` (Configuration.java:831) allows any authenticated client to set any property with no validation, an attacker can inject shell metacharacters (`;`, `$()`) into these properties to execute arbitrary commands.

**Note:** Other code paths in `LinuxUtils.java` (gateway, SSID, WiFi key) use `exec2(String)` which calls `Runtime.exec(String)` — this tokenizes via `StringTokenizer` without invoking a shell, so those are NOT vulnerable to shell metacharacter injection.

## Changes

- **`IOUtils.java`**: Replaced `sh -c "mount | grep"` with running `mount -t nfs` via array exec and performing the case-insensitive match in Java. Eliminates the shell entirely.
- **`MediaFile.java`**: Replaced `sh -c` concatenation with the `String[]` array form of `exec2()`, which passes arguments directly to the OS without shell interpretation.

### DIFF
--- a/java/sage/IOUtils.java
+++ b/java/sage/IOUtils.java
@@ -1143,8 +1143,9 @@ public class IOUtils
     if (!Sage.LINUX_OS) return NFS_MOUNT_FAILED;
     if (nfsPath.startsWith("nfs://"))
       nfsPath = nfsPath.substring(6);
-    // Check if the mount is already done
-    if (IOUtils.exec2(new String[] { "sh", "-c", "mount -t nfs | grep -i \"" + localPath  + "\"" }) == 0)
+    // Check if the mount is already done — avoid sh -c to prevent shell injection via localPath
+    String mountOutput = IOUtils.exec(new String[] { "mount", "-t", "nfs" });
+    if (mountOutput != null && mountOutput.toLowerCase().contains(localPath.toLowerCase()))
     {
       //if (Sage.DBG) System.out.println("NFS Mount already exists");
       return NFS_MOUNT_EXISTS;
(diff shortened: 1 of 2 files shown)


Narflex:
This one looks fine. :)  Although these code paths are never executed AFAIK, they were from things we never actually released (but they were functional, so there may be some way to enable them).

r3352:
Thanks! And good context — I noticed those code paths seemed like they might be dormant. Even if they're not actively triggered today, the fix is cheap insurance in case someone does find a way to enable them (or if they get wired up in a future change). Appreciate you reviewing these.

Narflex:
In thinking about this more...I may have used 'sh' to execute these because of path issues.

Do you have a Linux version running in a desktop environment where you can check to see if simply ExecuteProcess("mount") (or whatever the right syntax is) actually works?

r3352:
I actually spun this up on a Linux box to check. Both `Runtime.exec(new String[]{"mount", "-t", "nfs"})` and `Runtime.exec(new String[]{"sh", "-c", "mount -t nfs"})` produce identical results. Java uses `execvp()` under the hood so it searches PATH the same way a shell would. `mount` resolved to `/usr/bin/mount` without any issues. So the array form should be a safe drop-in replacement without PATH concerns.

r3352:
Reference - 503778570

